PR #2, opened 2015-07-23
Shortcutting the process

martinthomson:
Following up on Ted's suggestion at the meeting.


hardie:
Shortcutting the process, per Martin's PR.

PR #6, opened 2015-09-27
Address signature reuse vulnerability

bifurcation:
@agwa pointed out on the IETF ACME list that the current ACME challenges are vulnerable to a signature reuse attack, with an artfully chosen account key pair. This PR addresses that vulnerability by replacing the signature with a digest and reducing the client's control over the contents of the validation object.

This PR is ported from letsencrypt/acme-spec#223 (and based on top of #5).  It also includes some additional changes to address comments in that PR.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1503,6 +1503,19 @@ validation of domain names.  If ACME is extended in the future to support other
 types of identifier, there will need to be new Challenge types, and they will
 need to specify which types of identifier they apply to.
 
+[[ Editor's Note: In pre-RFC versions of this specification, challenges are
+labeled by type, and with the version of the draft in which they were
+introduced.  For example, if an HTTP challenge were introduced in version -03
+and a breaking change made in version -05, then there would be a challenge
+labeled "http-03" and one labeled "http-05" -- but not one labeled "http-04",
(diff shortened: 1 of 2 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1663,38 +1656,32 @@ server at the domain such that when a handshake is initiated with the Server
...
 of the domain by verifying that the TLS server was configured appropriately.
 
-1. Verify the validation JWS using the account key for which the challenge
-   was issued.
-2. Decode the payload of the JWS as UTF-8 encoded JSON.
-3. Verify that there are exactly two fields in the decoded object, and that:
-  * The "type" field is set to "dvsni"
-  * The "token" field matches the "token" value in the challenge
-4. Open a TLS connection to the domain name being validated on port 443,
+1. Verify that the "token" value in the response matches the "token" field in
```
jsha: What does this verification do? Why not just have the client POST an empty body to the challenge URI, or a body with some indicator that it is ready to proceed?

bifurcation: As discussed on the list, this check helps ensure that an intermediary didn't provide a bogus authorized key object.

Also, please raise these issues on the mailing list rather than using the PR thread.

PR #7, opened 2015-09-27
Address default virtual host risks

bifurcation:
On the mailing list, @pde and @jdkasten pointed out some risks arising from the default virtual host behavior of some web servers.  This PR integrates the changes they suggest, namely (1) removing the "tls" option from the `simpleHttp` challenge, and (2) adding iterations to the `dvsni` challenge.

This PR is based on top of #5 and #6.

### DIFF
--- a/draft-ietf-acme.md
+++ b/draft-ietf-acme.md
@@ -1179,7 +1179,6 @@ Host: example.com
 {
   "resource": "challenge",
   "type": "simpleHttp",
-  "tls": false
 }
 /* Signed as JWS */
 ~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


bluecmd:
Sorry, feel free to tell me to go away, but reading the backing decision to drop HTTPS for SimpleHTTP doesn't make sense to me.

```
 - SimpleHTTP is vulnerable if no default vhost is set explicitly, the
   server accepts the "tls": true parameter as currently specified, and
   the domains served over TLS/HTTPS are a strict subset of those served over
   HTTP.
```

If I understand everything correctly, you control the originator of the HTTPS requests and the requests are always sent to domains. So why can't you just enforce to always send the Host header and this wouldn't be an issue?

pcworld:
From [the original mail](https://mailarchive.ietf.org/arch/msg/acme/B9vhPSMm9tcNoPrTE_LNhnt0d8U):

> Apache (and I gather nginx too) have the subtle and non-intuitive
behaviour that if a default TLS/HTTPS virtual host is not configured
explicitly, one will be selected based on the ordering of vhosts in the
webserver configuration (in practice, this often amounts to alphabetical
ordering of files in a configuration directory).

As far as I know Apache does the exact same thing for nonencrypted HTTP hosts (this is also acknowledged by a later mail in the thread), so the reasoning is a little weird. I suppose the idea is that hosters are more likely to have HTTP configured correctly rather than HTTPS?

PR #8, opened 2015-09-27
Add explicit versioning to challenges

bifurcation:
The WG is presumably going to want to iterate on challenges as work on this spec progresses.  In order to enable implementations of pre-RFC versions of the spec to smoothly upgrade, this PR introduces explicit versioning on challenge names.  Each challenge name is now of the form "type-version", where "type" is the general type of challenge (HTTP, SNI, DNS, PoP), and "version" is the version of the draft that last changed the challenge.  For example, the version of the HTTP challenge in this draft is "http-00".

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1144,12 +1144,12 @@ Link: <https://example.com/acme/new-cert>;rel="next"
     {
       "type": "http-01",
       "uri": "https://example.com/authz/asdf/0",
-      "token": "IlirfxKKXAsHtmzK29Pj8A"
+      "authorizedKey": "SXQe-2XODaDxNRsbp0h...fMsNxvb29HhjjLPSggwiE"
     },
     {
       "type": "dns-01",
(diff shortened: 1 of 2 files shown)


martinthomson:
This has a lot of other changes in it.

I don't know if you value editorial convenience, but the current approach is hard to maintain.  In HTTP/2, we had a version identifiers section up front of the document.  That section explained that identifiers had the draft suffix appended and that the examples all omitted this suffix for convenience.

This protocol is different because you see the identifiers all over and there are things like signatures that cover them.  If you are OK with changing the identifers (all of these should be changed to -01 any way), then you should be OK.  I would add a preprocessing step, but that's me.

bifurcation:
Yeah, I see what you're saying, @martinthomson, but I think I prefer to be explicit here.  There are more version identifiers hanging around.

pde:
In favour of adding numbering to the Challenges.  @jdkasten do you prefer DVSNI or TLS SNI as a name?

PR #18, opened 2015-10-15
Clarify encoding for certs in PoP challenge

jsha:
They should be base64url rather than base64, for symmetry with rest of spec.


bifurcation:
Don't think this is totally necessary, since we say at the front that when we say "base64", we mean the special variant.  But happy to be clearer here.

PR #19, opened 2015-10-17
Update template files

martinthomson:
I've made some major changes to how this template thing works.  It's much nicer now, I think.

Note: the old CI stuff leaked keys, you will need to remove the old API key.  Add a new one to [circle CI](https://circleci.com), which I think is better.


richsalz:
lib/main.mk not found?

martinthomson:
You need to run `git submodule update--init` to get things started.

martinthomson:
Oh, and now that you have this, someone with admin rights to the repo should re-enable circle (or travis) builds and add a github token to the environment variables in the config so that the editor's copy remains up-to-date.  I can't do that for you, instructions are here: https://github.com/martinthomson/i-d-template#automatic-update-for-editors-copy

@bifurcation, @richsalz, @hardie all should be able to do that.

hardie:
On Sun, Oct 18, 2015 at 8:14 PM, Martin Thomson <notifications@github.com>
wrote:

> Oh, and now that you have this, someone with admin rights to the repo
> should re-enable circle (or travis)
>
​Any particular reason to prefer one over the other?  I tend to update
manually, as not every change is worth an editors copy, but I'm happy to
shift to either one, if that's the preference.

Ted​

martinthomson:
> ​Any particular reason to prefer one over the other?

I have a mild preference for circle, since it seems a little cleaner, architecturally speaking.  Feature-wise, they are basically identical.  Circle might be slightly faster.

richsalz:
I agree with @hardie, no need to rebuild every change and am fine with manual.
did the init on the submodule branch, restored issues.json and pushed; all is happy now.

PR #21, opened 2015-10-20
Updating makefile and template

martinthomson:
This avoids the problem that @richsalz had with submodules.  Also, I've updated the template a little.


martinthomson:
(to test, remove the lib directory and build again)

PR #22, opened 2015-10-29
Simplify TLS SNI challenge

jsha:
Simplify the TLS SNI challenge by removing the iterations concept, which I believe is not necessary in practice and complicates the protocol needlessly.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1645,35 +1645,26 @@ token (required, string):
 at least 128 bits of entropy, in order to prevent an attacker from guessing it.
 It MUST NOT contain any characters outside the URL-safe Base64 alphabet.
 
-n (required, number):
-: Number of tls-sni-01 iterations
-
 ~~~~~~~~~~
 {
   "type": "tls-sni-01",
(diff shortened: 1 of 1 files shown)


kelunik:
Implementing TLS SNI is currently really complicated, yes.

kuba:
FYI, https://github.com/letsencrypt/letsencrypt/tree/v0.0.0.dev20151108 implements protocol according to the contents of this PR - i.e. without iterations.

kuba:
(so does current Boulder)

bifurcation:
I think this will be OK to land once #54 lands.  

We should consider whether we want to bump the version number in the challenge.  Principle would dictate that we should, since this breaks compatibility with what was published in -01.  But since I'm not aware of any implementation of "tls-sni-01" as specified in -01 (vs. the off-spec implementation in boulder), I'm inclined to just leave it.

mholt:
FWIW on the client side of things, xenolf/lego did not implement the iterations for the tls-sni-01 challenge.

hlandau:
https://github.com/hlandau/acme implements iterations but takes the absence of an "n" parameter to be equivalent to an "n" parameter of 1, which is compatible.

bifurcation:
Great, I'm taking this as sufficient evidence that a version bump is not needed :)

PR #35, opened 2015-11-12
Fix typo

irons:
Xcode is spelled with [one capital letter](https://developer.apple.com/xcode/).


bifurcation:
This is a result of the [template](https://github.com/martinthomson/i-d-template) from which this repository was created.  Please file an issue there, and we'll get the change here once it's fixed upstream.

PR #39, opened 2015-11-12
Add roll-over for account keys

### LINKED ISSUE #14: Support key rollover for account keys
Account keys should change over time to improve security.  We should have a way to change the key associated with an account with changing anything else about the account.

Something liike:

Post: /rotate-key
{
"key":  { /* JWK key */ }
}

Note: The request should be signed with the old key.  All authorizations should remain with the account.  If there are any requests being validated then they should still use the old key (as specified by the oriignal authorization request.

bifurcation:
Fixes #14

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -912,6 +912,57 @@ Host: example.com
 /* Signed as JWS */
 ~~~~~~~~~~
 
+## Account Key Roll-over
+
+A client may wish to change the public key that is associated with a
+registration, e.g., in order to mitigate the risk of key compromise.  To do
+this, the client first constructs a JSON object representing a request to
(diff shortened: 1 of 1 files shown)


jsha:
What advantage does including the old key provide?

bifurcation:
Having the new key sign over the `oldKey` field makes the hand-over mutual -- the old key blesses the new key by signing over the `newKey` JWS and the new key accepts by signing over the `oldKey` thumbprint.  Basically, the `oldKey` field is acting as an identifier for the registration for purposes of the new key's acceptance of the hand-over.  Note that this is well-defined because the server is required to only have one registration for a given key-pair (it MUST return a 409/Conflict if there's an attempt to duplicate).

PR #41, opened 2015-11-12
Remove MAC-based account recovery

bifurcation:
As discussed at the F2F meeting at IETF94, MAC-based account recovery is not adding a lot of value, and it adds significant complexity to the spec.  (And nobody has implemented it.)  This PR removes MAC-based recovery and the crypto that it depends on.


bifurcation:
Seeing no complaints on the mailing list, mergining.

PR #46, opened 2015-11-16
Minor editorial changes

werksjan:
editorial change


hardie:
These LGTM as editorial.  I note that the inclusion of "revoked" in https://github.com/werksjan/acme/commit/fbccb2ea3d22c98f0a4b9ab5d079455b64bb8272 is on the edge, but I think it was intended and could be merged as a result.

werksjan:
I did think the mentioned change might be an edge case but came to the same conclusion. Happy to be of any help.

bifurcation:
LGTM as well.  Thanks @werksjan.

PR #51, opened 2015-12-07
Remove one gendered pronoun

lsb:
Avoid changing the "him" in the MitM paragraph, because of the inherent notion of a *Man* in the Middle, which is a larger discussion.


bifurcation:
LGTM.  Thanks for the elegant fix.

PR #52, opened 2015-12-15
Agreement URI MUST be updated when terms change

kelunik:
I'm not sure whether we should add something like:

```
For any other changes, e.g. typographical changes, the URI SHOULD NOT change.
```


mholt:
I think "in a way that requires an agreement update" is clear enough. Users won't need to re-agree if just fixing a typographical error.

PR #56, opened 2015-12-28
Change 'proofOfPossession' to 'proof-of-possession'

### LINKED ISSUE #53: Inconsistent naming convention of the challenges
Any reason for the inconsistent naming of the challenge types?

Specifically `tls-sni-01` vs `proofOfPossession-01`

Wouldn't it make sense to stick to dash delimited words? or just camelCase, not that it matter as long as its consistent.

I would personally avoid camelcase in that context since there is already a lot acronyms.

bifurcation:
Fixes #53


unixcharles:
awesome thanks :+1:

PR #65, opened 2016-01-06
Register invalidEmail error type.

jsha:


bifurcation:
Note for later: It would be better for this to be more generic, e.g., `invalidContact`.

PR #69, opened 2016-01-21
Add operational guidance for DNSSEC

hlandau:


jsha:
I approve of this change.

bifurcation:
This looks good to me.  I would like to tighten up the language a bit to clarify that this applies to validation of DNS identifiers, but will do that in a follow-on PR.

PR #71, opened 2016-01-22
Change TLS-SNI to require hostname not specified by validator

hlandau:
This replaces tls-sni-01 with tls-sni-02, which requires that domains
being validated return a certificate which contains a hostname that is
not revealed by the server in its handshake. This ensures that
challenges cannot be passed by domains which auto-generate self-signed
certificates in response to requests with SANs for domains not
previously seen.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1584,7 +1584,7 @@ form. The dNSName is "x.y.ka.acme.invalid" where x is the first half of the
 hexadecimal representation and y is the second half.
 
 The client MUST ensure that the certificate is served to TLS connections
-specifying a Server Name Indication (SNI) value of SAN B.
+specifying a Server Name Indication (SNI) value of SAN A.
 
 The response to the TLS-SNI challenge simply acknowledges that the client is
 ready to fulfill this challenge.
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1603,14 +1609,19 @@ Given a Challenge/Response pair, the ACME server verifies the client's control
...
 2. Open a TLS connection to the domain name being validated on the requested
-   port, presenting the value "\<Z[0:32]\>.\<Z[32:64]\>.acme.invalid" in the
-   SNI field (where the comparison is case-insensitive).
-3. Verify that the certificate contains a subjectAltName extension with the
-   dNSName of "\<Z[0:32]\>.\<Z[32:64]\>.acme.invalid", and that no other dNSName
-   entries of the form "*.acme.invalid" are present in the subjectAltName
-   extension.
+   port, presenting SAN B in the SNI field.
+
+   The server MUST request the certificate using an SNI value of SAN B.
```
jehiah: I think this is backwards. `x.y.token.acme.invalid` should be used for the request, ie `SAN A`, and `SAN B` (the key authorization) should not be revealed.

hlandau: SAN B doesn't reveal the key authorization (and therefore the token) because it is a hash of it.

jehiah: True, i believe either order is sufficient for security, but I think it's desirable (helpful?) to keep with the pattern of `http-01` where the token is used in the request, and the key authorization is provided in the response.

draft-ietf-acme-acme.md
```diff
@@ -1560,28 +1560,34 @@ It MUST NOT contain any characters outside the URL-safe Base64 alphabet.
...
+dNSNames.
+
+SAN A MUST be constructed as follows: compute the SHA-256 digest of the
+UTF-8-encoded challenge token and encode it in lowercase hexadecimal form.  The
+dNSName is "x.y.token.acme.invalid", where x is the first half of the
+hexadecimal representation and y is the second half.
+
+SAN B MUST be constructed as follows: compute the SHA-256 digest of
+the UTF-8 encoded key authorization and encode it in lowercase hexadecimal
+form. The dNSName is "x.y.ka.acme.invalid" where x is the first half of the
```
kelunik: How about `x.y.key.acme.invalid`?

PR #72, opened 2016-01-24
Add metadata to directory

### LINKED ISSUE #59: Add terms-of-service to directory
Right now, in order to discover the CA's subscriber agreement, you have to create a new registration and get it back in a Link header.  Since this is a static property of the CA, it should be possible to get it without taking any action.  The directory is the natural place for it.

hlandau:
This commit does two things: Firstly, it adds a metadata section to the
directory, allowing terms of service and website URL to be stored, as
well as a list of recognised CAA issuer identifiers.

Secondly, it adds a Link (rel="directory") to all resources pointing
back to the directory. This ensures that given the URL of an ACME
resource, the URL of the directory needed to perform operations relating
to it can be found.

Fixes #59.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -43,6 +43,7 @@ normative:
   RFC5988:
   RFC6066:
   RFC6570:
+  RFC6844:
   RFC6962:
   RFC7159:
   RFC7469:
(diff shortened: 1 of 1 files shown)


bifurcation:
This got merged in d34995e, but Github is not recognizing it because I used the wrong commit message.

PR #74, opened 2016-01-27
Wildcard support

### LINKED ISSUE #73: Support authorizations for wildcard names

hlandau:
Fixes #73.

See also https://github.com/letsencrypt/boulder/issues/21

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -481,9 +481,12 @@ client completes all challenges.
 
 The only type of identifier defined by this specification is a fully-qualified
 domain name (type: "dns").  The value of the identifier MUST be the ASCII
-representation of the domain name.  Wildcard domain names (with "*" as the first
-label) MUST NOT be included in authorization requests.  See
-{{certificate-issuance}} below for more information about wildcard domains.
+representation of the domain name.  A wildcard domain (a domain starting with
+one or more `*` labels) MAY be the subject of an authorization request. A
+partial-match wildcard domain (a domain which contains a label which contains
+`*` but is not `*`, for example `foo*.example.com`) MUST NOT be included in
+authorization requests. See {{certificate-issuance}} below for more information
+about wildcard domains.
 
 ~~~~~~~~~~
 {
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1185,12 +1188,16 @@ key that signed the request.
...
-allow the client to issue a certificate with an iPAddress subjectAltName, even
-if it contains an IP address to which example.com resolves.
+client's account key. Servers SHOULD NOT extend authorization across identifier
+types.  For example, if a client is authorized for "example.com", then the
+server should not allow the client to issue a certificate with an iPAddress
+subjectAltName, even if it contains an IP address to which example.com
+resolves.
+
+Wildcard names SHOULD NOT be accepted unless a client is authorized for that
+exact wildcard name. For example, an authorization for `*.*.example.com`
```
pzb: Only wildcards in the form `"*."` plus a valid FQDN should be supported.  TLS clients generally do not support multiple asterisk characters in a name nor do they support partial label wildcards (e.g. f*.example.com, *g.example.com, or f*g.example.com).

draft-ietf-acme-acme.md
```diff
@@ -1428,6 +1436,35 @@ over the same type, e.g., ["http-03", "http-05"].  Challenges within a type are
...
+
+When validating a challenge other than a proof of possesion challenge which is
+part of an authorization for a wildcard name (a domain name which contains one
+or more `*` labels), the challenge is validated as follows:
+
+- A CA chooses some positive integer n by policy.
+
+- The CA generates n random strings for each of the m `*` labels. Let `s(i,j)`
+  denote string i for `*`-label j. Each string MUST consist of lowercase
+  letters, decimal digits and hyphens and MUST NOT exceed 63 characters. Each
```
pzb: This should say that it MUST be a valid DNS label and SHOULD not start with `xn--`.  Pure random generation could end up with starting or ending with `-` which is not valid and could end up looking like a punycode label.

draft-ietf-acme-acme.md
```diff
@@ -1428,6 +1436,35 @@ over the same type, e.g., ["http-03", "http-05"].  Challenges within a type are
 testing the same capability of the domain owner, and it may not be possible to
 satisfy both at once. ]]
 
+## Validation of Wildcard Authorizations
```
pzb: This should be a new challenge type or there should be some flag to indicate that this challenge method was used with an existing type.

draft-ietf-acme-acme.md
bifurcation: I'm hoping that this is a typo, and he actually meant `*.example.com`?

draft-ietf-acme-acme.md
bifurcation: I agree that this should be a different challenge type.

draft-ietf-acme-acme.md
hlandau: This can/should be left as a policy consideration for a particular CA.

draft-ietf-acme-acme.md
hlandau: So you want `wildcard-http-01`, `wildcard-tls-sni-01` and `wildcard-dns-01`? That seems excessive. Adding `"wildcard":true` to the authorization object seems preferable.

fpietrosanti:
The Tor2web project would to use it and implement/integrate this specs from letsencrypt implementation as soon as it's available https://github.com/globaleaks/tor2web

WinPack:
So, @bifurcation how is discussion progressing? Has there been any decision? If not, when approximately will there be?

richsalz:
All discussion happens on acme@ietf.org; please join and participate there.  See https://www.ietf.org/mailman/listinfo/acme to join.
There has been almost no discussion.

bifurcation:
After discussion on-list and at IETF 95, closing this in favor of #124

PR #75, opened 2016-01-28
suggest using DNS over TCP when verifying challenges

mozfreddyb:
I've noticed that boulder uses DNS over TCP. I suppose this has not been an arbitrary choice, but something that would be useful to document.


jsha:
As an FYI, Boulder currently uses TCP for DNS to its local authoritative resolver, but the resolver uses UDP. We tried making the resolver do upstream resolution with TCP but got a large number of failures.

PR #76, opened 2016-01-28
Considerations for Domain Holders

mozfreddyb:
A dynamic DNS provider might allow users to request TXT records for `_acme-challenge.example.tld`.
This is mitigated by placing domains on the public suffix list.

I have also noted that the public suffix list is just mentioned on the side in this draft. Should I follow up with another pull request that makes the PSL a reference pointing to https://publicsuffix.org/?


hlandau:
Discussion should be done on the mailing list.

jsha:
This was discussed on the mailing list and we didn't conclude a change needed to be made. We can bring it up again if you feel differently @mozfreddyb. Thanks!

PR #78, opened 2016-01-31
Make authorizations and certificates mandatory for registrations

kelunik:
https://mailarchive.ietf.org/arch/msg/acme/M87w3UsOan2kmZdW0yuoLgG7u6c and following messages.


bifurcation:
This is fine semantically, but the convention here has been to use "(required, string)" rather than just "string".  I will clean up and merge.

PR #82, opened 2016-02-09
Allow the client to give a preference for a specific IP in http-01

horazont:


bifurcation:
This looks fine to me, and consistent with the list discussions.  (I have a couple of minor issues, which I filed as #90.)  I am merging this in preference to #83, which got negative feedback on the list.

PR #83, opened 2016-02-09
http-01: Use client's SRV records for http validation

micw:


draft-ietf-acme-acme.md
```diff
@@ -1462,6 +1462,20 @@ control over a domain name by proving that it can provision resources on an HTTP
 server that responds for that domain name.  The ACME server challenges the
 client to provision a file with a specific JWS as its contents.
 
+The server will first check if there exists an SRV-Record
+"_acme-challenge._http-01._tcp." followed by the domain name being validated.
```
hlandau: I think this should really be something like `_acme-challenge-http._tcp`.

rhansen:
:+1:

bifurcation:
Closing in preference of #82 as a solution for the servers-not-sync'ed problem.

PR #85, opened 2016-02-22
CAA account key binding support

hlandau:
This specifies a standard format of CAA record parameter to nominate an ACME account key thumbprint.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -43,6 +43,7 @@ normative:
   RFC5988:
   RFC6066:
   RFC6570:
+  RFC6844:
   RFC6962:
   RFC7159:
   RFC7469:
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1801,6 +1802,39 @@ If all of the above verifications succeed, then the validation is successful.
 If no DNS record is found, or DNS record and response payload do not pass these
 checks, then the validation fails.
 
+# Use with CAA Records
+
+An ACME server SHOULD support CAA DNS records as described in {{RFC6844}}. The
+server SHOULD look for such records when issuing authorizations, as opposed to
+when issuing certificates.
```
jcjones: Since authorizations may last a long time, maybe the implementing CA should check at both authorization time and (eventual) issuance time.

hlandau: I do wonder about the use case for long authorizations. Since the whole purpose of ACME is to automate issuance, an authorization need only be long enough for an ACME client to conclude the immediate issuance transactions it needs to make...?

bifurcation:
Rejecting this PR because the account-key binding should really go in another document.  FWIW, I do think having an account-key binding for CAA is a good thing, but it's separable.

@hlandau, if you want to submit a separate PR with just the recommendation for doing CAA, I would be happy to merge it.  I agree with @jcjones, though, that it might also be advisable to check at cert issuance time.  Maybe "SHOULD check at authz time, MAY also check at issuance time, especially if they're far apart"?

PR #86, opened 2016-03-01
Authorization / certificate lists can be paged

jsha:


bifurcation:
This looks consistent with list discussion.  Thanks.

PR #101, opened 2016-03-21
Remove proof-of-possession challenge

bifurcation:
There has been no implementer interest in the proof-of-possession challenge, and it makes the document more complex by having a completely different structure than the other challenges.  If it is decided later that such a challenge is necessary, it can be added in a follow-on extension specificiation.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1502,22 +1502,8 @@ help the client diagnose why they failed the challenge.
 Different challenges allow the server to obtain proof of different aspects of
 control over an identifier.  In some challenges, like HTTP and TLS SNI, the
 client directly proves its ability to do certain things related to the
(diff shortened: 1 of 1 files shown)


jcjones:
Generally this seems fine,  and I wholeheartedly agree with the purpose of the removal (for now). One problem I can't open a comment for: Line 2099, the last paragraph before "Preventing Authorization Hijacking", still refers to the PoP Challenge. Without the PoP Challenge wording, a new reader may take that this to be referring to the implicit Proof of Possession of the ACME Account Key and thus be confused.

Since it's non-normative, perhaps you could simply modify that paragraph into something more like:

> An active attacker on the validation channel can subvert the ACME process, by performing normal ACME transactions and providing a validation response for his own account key. The risks due to hosting providers noted above are a particular case. For identifiers where the server already has some credential associated with the domain this attack can be prevented by requiring the client to **utilize the previously-registered account key**.

PR #102, opened 2016-03-21
Replace in-band account recovery with `meta`

bifurcation:
We already removed token-based account recovery; this PR removes contact-based recovery.  It was already almost entirely unspecified, due to it being a process that primarily takes place outside of ACME.  Removing account recovery from ACME simplifies both the protocol and the security considerations.

As an alternative, this PR adds an `account-mgmt` field to the `meta` dictionary in the directory, which the CA can use to expose account management functions such as recovery.  But you can also imagine using this for more sophisticated account management functions, like associating an ACME account with some other account in a different system.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -705,21 +705,20 @@ the service provided by the ACME server.
 
 The following metadata items are defined, all of which are OPTIONAL:
 
-"terms-of-service":
-: A string containing a URI identifying the current terms of service.
+"terms-of-service" (optional, string):
+: A URI identifying the current terms of service.
 
-"website":
-: String containing an HTTP or HTTPS URL locating a website providing more
-information about the ACME server.
+"website" (optional, string):
+: An HTTP or HTTPS URL locating a website providing more information about the
+ACME server.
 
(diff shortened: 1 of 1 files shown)


bifurcation:
(Also threw in some editorial corrections, sorry)

jcjones:
The editorial corrections are good and IMO should be taken in any circumstances.

I agree this change does what it says it does, so r+.

One concern, which I will cross-post to the list, is that the placement of this URL in the directory means that it is not customizable; It could be a useful thing to, after a signed request, give back a URI that contains a one-time value that would help tie a browser session to a given account key. I don't see a good endpoint for such a thing at present, so the hour is probably too late for such a change.

bifurcation:
@jcjones I agree that could be useful, but I can think of other, simpler ways to provide it (e.g., a link relation from a registration URI).  If you think it's important, please file a follow-on issue.

jsha:
@bifurcation: See my feedback on the ACME list. I disagree that this should be merged now. Could you please submit a revert and seek further discussion?

bifurcation:
Sorry, I misunderstood.  Will revert.

PR #107, opened 2016-03-21
Follow type convention for "meta" fields

bifurcation:
(Re-adding the editorial changes that got lumped in #102)


jcjones:
r+; thanks for pulling these out alone

PR #108, opened 2016-03-21
Remove account recovery

bifurcation:
This PR is a re-do of #102.  It only removes recovery; it does not add the "account-mgmt" metadata item.


jsha:
LGTM

PR #122, opened 2016-05-03
Account compromise mitigations

sheelx86:
Based on the mailing list thread "[Acme] [acme] Account deletion for security currently useless if rolled over".

Short reasons: Not too long ago, the possibility for deleting accounts was introduced. Partially this, partially the account key rollover, has some problems in case the account key gets lost/stolen/anything. 

Basically, this is for preventing the attacker from deleting/rekeying/... the account to prevent the rightful owner from revoking/deleting the new "evil" certificates.

Sorry, it got quite a lot.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -496,8 +496,9 @@ indicate why the server was unable to validate authorization.
 In this section, we describe the certificate management functions that ACME
 enables:
 
-  * Account Key Registration
-  * Account Key Authorization
+  * Account Registration
+  * Account Authorization
+  * Account Deletion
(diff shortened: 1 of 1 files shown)


jsha:
We wound up landing a lot of similar concepts in https://github.com/ietf-wg-acme/acme/pull/125. I think the big part that you've proposed here, keeping a history of keys and allowing older keys to still revoke certs and authorizations, was discussed on-list and we thought that it would be better to automatically deactivate old authorizations when new ones are validated. If you think there's more work to be done, can you re-propose based on the current version and post to the list?

Thanks,
Jacob

PR #124, opened 2016-05-06
Enable the server to specify requirements for issuance

### LINKED ISSUE #115: Cashbox question
At IETF 95, issue came up that supporting pre-ACME CAs would likely involve the need for payment.  This is a stub issue to start discussion of how ACME might respond.

bifurcation:
Fixes #115 
(Cash-box can be implemented as an out-of-band precondition)

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -208,7 +208,8 @@ The first phase of ACME is for the client to register with the ACME server.  The
 client generates an asymmetric key pair and associates this key pair with a set
 of contact information by signing the contact information.  The server
 acknowledges the registration by replying with a registration object echoing the
-client's input.
+client's input.  The server can also provide terms of service at this stage,
+which the client can present to a human user.
 
 ~~~~~~~~~~
       Client                                                  Server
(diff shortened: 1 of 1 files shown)


kelunik:
I'd prefer to have the missing identifier in the precondition response instead of an already created and pending authorization.

draft-ietf-acme-acme.md
```diff
@@ -1207,9 +1315,14 @@ authorized).
 
 It is up to the server's local policy to decide which names are acceptable in a
 certificate, given the authorizations that the server associates with the
-client's account key.  A server MAY consider a client authorized for a wildcard
-domain if it is authorized for the underlying domain name (without the "*"
-label).  Servers SHOULD NOT extend authorization across identifier types.  For
+client's account key.  For example, a server might have a policy that it will
+consider a client authorized for a wildcard domain if it is authorized for some
+set of names that match the wildcard.  If the client does not yet have
```
joshtriplett: It seems dangerous to hint at the right requirements without actually stating them explicitly or providing any guidance for prospective server implementers.  Getting these requirements wrong would allow the incorrect issuance of wildcard certificates.

For example, it doesn't suffice to generate N random subdomains and require that the client match those, without also requiring that the same client (with the same key) prove ownership of the containing domain (e.g. proving ownership of example.org in addition to random examples of *.example.org).  Otherwise, anyone could obtain a wildcard certificate for any site that allows user-created subdomains with arbitrary names and user-controlled content, by manually creating the requested names.  Requiring proof of ownership of the

PR #125, opened 2016-05-12
Replace deletion with deactivation.

jsha:
This simplifies both the spec and the implementation, and does not attempt to
impose specific data retention policy on the CA.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -593,6 +593,9 @@ key (required, dictionary):
 : The public key of the account key pair, encoded as a JSON Web Key object
 {{RFC7517}}.
 
+status (required, string):
+: "good" or "deactivated"
+
 contact (optional, array of string):
 : An array of URIs that the server can use to contact the client for issues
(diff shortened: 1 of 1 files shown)


WinPack:
I don't really agree about that one. Why not delete completely the data instead of deactivating? Why shouldn't a user be able to request full deletion and no retention of any of their account data?

jsha:
Nothing in this change prevents CAs from choosing to delete account data immediately upon deactivation, but the spec should not try to dictate policy.

bifurcation:
I think this PR is fine in principal, but I have some minor concerns:

* If we're going to use the `status` field, we should prefer to use the values that are already there, namely `valid` and `invalid`.  Do you think there's a semantic difference between that and what you're proposing?

* Your edits remove the guidance on what the server should return from these requests.  Please add that back in.

* It occurs to me looking at this that the first version of authz deletion should have specified that the server MUST verify that the account is the owner of the authz before allowing it to deactivate the authz.  Could you please add that while you're editing here?

jsha:
Done.

PR #133, opened 2016-06-03
Add a port argument to http-01 and tls-sni-02.

coderanger:
This allows running challenges without explicit coordination from the process(s) on port 80 or 443, but still requires the user running the challenge to have root-level networking permissions.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1544,6 +1544,10 @@ If given, the server SHOULD connect to that specific IP address instead of
 arbitrarily choosing an IP from the set of A and AAAA records to which the
 domain name resolves.
 
+port (optional, integer):
+: A TCP port between 1 and 1023 (inclusive). If given, the server MUST connect
+on this port instead of the default TCP port (80).
+
 ~~~~~~~~~~
 /* BEGIN JWS-signed content */
 {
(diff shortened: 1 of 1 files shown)


coderanger:
I've tried emailing the list twice now, but I think its getting held in a moderation queue. Can someone poke the list?

richsalz:
There is nothing in the moderation queue.  Please email acme-chairs at ietf.org if you are having problems.

jsha:
@coderanger: Can you try posting again? I'd like to discuss the question of alternate ports on-list. Closing this pull request for now as part of a cleanup. We can reopen, or create a new one, once there's a conversation on the list.

FWIW, I had some problems recently with my mail winding up in people's spam because of DMARC settings on my sending domain.

coderanger:
I was corresponding with @hardie a bit and I think he was unable to see any obvious reasons me emails aren't getting through, but I'll try poking DKIM with a sharp stick :)

coderanger:
I've confirmed my DKIM records (and SPF) are correct. Will try emailing the list again for old times sake I guess?

richsalz:
Try.

coderanger:
Yep, I just sent it again.

coderanger:
Doesn't appear to have made it through, either with or without the GitHub URL.

PR #137, opened 2016-06-09
Revert "suggest using DNS over TCP"

jsha:
This reverts commit 556ab4e0f3ff4d402bea530dee7c22ca9c3306db (pull #75).

Deployment experience has shown that DNS via TCP to remote hosts is too unreliable to use in practice. Let's Encrypt uses DNS via TCP only within its own datacenter, and may stop doing that soon too.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -2127,10 +2127,6 @@ perform all necessary checks before issuing.
 
 There are certain factors that arise in operational reality that operators of
 ACME-based CAs will need to keep in mind when configuring their services.
-For example:
-
-* It is advisable to perform DNS queries via TCP to mitigate DNS forgery
-  attacks over UDP
 
 [[ TODO: Other operational considerations ]]
 


raybellis:
This seems short-sighted.  IMHO (and as the author of RFC 5966 and co-author of its update - RFC 7766) there's no good reason why _any_ authoritative server should be blocking TCP queries, and they SHOULD be tried first and used if responding.

jsha:
Fair enough. Do you have a recommended DNS server with that behavior, or a recommendation for how to configure Unbound (which Let's Encrypt uses) to do so? AFAICT Unbound will always try UDP first.

raybellis:
Unbound has "tcp-upstream: yes | no" but I suspect that might be an issue if it actually disables UDP.

Is there any good reason you're going via a local recursive server instead of talking directly to the user's authoritative server directly from the back-end code   [ having found the NS records via the recursor first, perhaps ].

A work around may be to use two (sets of) local DNS servers, one configured for TCP only, and the other using UDP, and failover on the client side if the former doesn't answer quickly enough (or can't because the far end doesn't support TCP).

jsha:
> Is there any good reason you're going via a local recursive server instead of talking directly to the user's authoritative server directly from the back-end code [ having found the NS records via the recursor first, perhaps ].

If we used UDP to find the authoritative server and then used TCP in Boulder to talk directly to that authoritative server, we wouldn't be getting the anti-spoofing benefits. Also, relying on Unbound for the full lookup means it can do CNAME following and DNSSEC validation for us.

> A work around may be to use two (sets of) local DNS servers, one configured for TCP only, and the other using UDP, and failover on the client side if the former doesn't answer quickly enough (or can't because the far end doesn't support TCP).

This is what I'm thinking of currently. However, in testing I'm finding the performance of Unbound with `tcp-upstream: yes` is dramatically lower than with `tcp-upstream: no`. I'm getting roughly 500 qps with TCP before I start getting large numbers of timeouts vs 3000 qps with UDP. I'm using a DigitalOcean box with 8 CPUs and 16 GB RAM, and this config (some access control and other info stripped). If you have any tips on how to improve the performance I'm all ears. Happy to take it to email or a phone call too if you have the time. Thanks!

```
server:
    tcp-upstream: yes
    outgoing-num-tcp: 10000
    outgoing-range: 10000
    so-rcvbuf: 8m
    so-sndbuf: 8m
    num-queries-per-thread: 5000
    verbosity: 0
    logfile: ""
    statistics-interval: 60
    extended-statistics: yes
    use-syslog: no
    log-time-ascii: yes
    chroot: ""
    username: "unbound"
    directory: "/etc/unbound"
    interface: 0.0.0.0
    interface: ::0
    port: 5053
    do-ip4: yes
    do-ip6: no
    do-udp: yes
    do-tcp: yes
    hide-identity: yes
    hide-version: yes
    harden-glue: yes
    harden-dnssec-stripped: yes
    use-caps-for-id: yes
    cache-min-ttl: 0
    cache-max-ttl: 60
    prefetch: no
    num-threads: 8
    incoming-num-tcp: 500
    msg-cache-slabs: 32
    rrset-cache-slabs: 32
    infra-cache-slabs: 32
    key-cache-slabs: 32
    rrset-cache-size: 256m
    msg-cache-size: 128m
    root-hints: "/var/lib/unbound/root.hints"
    unwanted-reply-threshold: 10000
    do-not-query-localhost: yes
    trusted-keys-file: /etc/unbound/keys.d/*.key
    auto-trust-anchor-file: "/var/lib/unbound/root.key"
    val-clean-additional: yes
```

jsha:
Closing this pending more research on feasibility of using tcp-upstream.

PR #138, opened 2016-06-09
Revert "Allow the client to give a preference for a specific IP in http-01"

jsha:
This reverts commit a5cb357a64ce5e46815544c9faaf94b97b3eba56

I think we did not reach consensus on the list about this feature.


lachesis:
Oh, it's a shame that this was removed. I was hoping to use this feature to help when my servers were behind a load balancer. We should probably remove this reference as well:
```
In addition, the client MAY advise the server at which IP the challenge is provisioned.
```

PR #145, opened 2016-07-08
Remove Flattened JSON Serialization requirement.

jsha:
The multiple signatures used for key rollover require non-flattened form.


bifurcation:
LGTM.

PR #150, opened 2016-07-08
Make key-change depend on both old and new keys

### LINKED ISSUE #149: Sign both the old key and the new key in roll-over
The current `key-change` mechanism has a potential copy/paste vulnerability.  Karthik proposed on the list:

> The safest design would be to have the JSON object include
> the thumbprints of both the old and new keys and then sign it
> with both the new and old keys.

That is, you need the full matrix of "X signs over Y" -- "new/old", "new/new", "old/old", "old/new" -- so that both sides explicitly agree to the (old, new) pair.

But the current design has only has "old/new" and "new/new".  That raises a risk that if you got an "old/new" signature, you could copy/paste the "new/new" signature and get a valid `key-change` request.

As long as we have parallel signatures (cf. #148), we need to have both the old and new keys in the payloads.

bifurcation:
Fixes #149

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -993,8 +993,7 @@ registration in order to recover from a key compromise or proactively mitigate
 the impact of an unnoticed key compromise.
 
 To change the key associate with an account, the client sends a POST request
-containing a key-change
-object with a "key" field containing a JWK representation of the new public key.
+containing a key-change object with the following fields:
 
 oldKey (required, JWS):
 : The JWS representation of the original key (i.e., the client's current account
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -992,8 +992,16 @@ A client may wish to change the public key that is associated with a
 registration in order to recover from a key compromise or proactively mitigate
 the impact of an unnoticed key compromise.
 
-To change the key associate with an account, the client POSTs a key-change
-object with a "key" field containing a JWK representation of the new public key.
+To change the key associate with an account, the client sends a POST request
+containing a key-change object with the following fields:
+
+oldKey (required, JWS):
```
jsha: s/JWS/JWK/ here and on the following four lines.

PR #154, opened 2016-07-08
Clarify the impact of SSRF

bifurcation:
@bburky does this look better?

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -2271,8 +2271,8 @@ the ACME server, even if it is not accessible to the ACME client.
 
 It might seem that the risk of SSRF through this channel is limited by the fact
 that the attacker can only control the domain of the URL, not the path.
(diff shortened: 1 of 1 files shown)


jsha:
LGTM, though I'd change the attacker from a "he" to a "they" on general principles.

bifurcation:
Not a huge fan of the singular "they", but can't be bothered to rewrite around it :)

bburky:
Yes, this is clearer. Thanks. Saying "will cause the server to query an arbitrary URI" is good. Technically you can even specify a non-HTTP protocol in an HTTP redirect, and this language should cover that too. (Go's net/http used by Boulder only does HTTP and HTTPS I think, but things like libcurl support a slightly insane number of protocols.)

I'm not sure if the usages of "URI" vs "URL" in this section are correct, you may want to make them consistent. I never know which one to actually use.

This looks good to me.

PR #156, opened 2016-07-08
Economize on nonces.

jsha:
Previously the server was required to provide a nonce on all successful
responses, including GETs. This makes certain nonce-storage techniques like an
in-memory list impractical, because the size of the list would have to scale
with GET requests rather than just authenticated POSTs.

This change reduces the scope of requests where nonces are required.

It also tweaks the example section for Replay-Nonce to not define the base64url
character set.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -350,22 +350,24 @@ requests have a mandatory anti-replay mechanism.  This mechanism is based on the
 server maintaining a list of nonces that it has issued to clients, and requiring
 any signed request from the client to carry such a nonce.
 
-An ACME server MUST include a Replay-Nonce header field in each successful
-response it provides to a client, with contents as specified below.  In
-particular, the ACME server MUST provide a Replay-Nonce header field in response
-to a HEAD request for any valid resource.  (This allows clients to easily obtain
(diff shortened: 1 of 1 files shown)


jdkasten:
This PR does encourage clients to store their most recent nonce. 
I have seen at least a few clients which fetch the directory before each submission for a fresh nonce. (Example: https://github.com/diafygi/letsencrypt-nosudo/blob/master/sign_csr.py#L26)

It seems unfortunate that with this PR, the only guaranteed way to begin the ACME process is by posting a failed registration request.  Clients may simply perform all of their transactions twice to get the same "nonce_req" result.

rolandshoemaker:
(Playing devils advocate) Why not just remove the nonce system entirely and recommend not terminating TLS at a CDN and/or recommend clients should pin API TLS keys?

jsha:
Both are great pieces of feedback! Can you post on the list? Most people there don't "watch" the repo.

rolandshoemaker:
Will do.

bifurcation:
Closing this pursuant to discussions at IETF 96.

https://www.ietf.org/proceedings/96/minutes/minutes-96-acme

jsha:
Reopening for further discussion on-list.

bifurcation:
Addressed by #181

PR #161, opened 2016-07-28
Drop OOB challenge

### LINKED ISSUE #152: Reduce the number of OOB things
We currently have two levels of out-of-band in the spec:

* Out-of-band requirements in the application object
* Out-of-band challenges

Should we get rid of the latter?  Or is it useful to let CAs migrate their authorization tracking to ACME but not their domain validation?

bifurcation:
Fixes #152


bifurcation:
Closing because some folks seem to think it is useful to have OOB at both levels.

PR #162, opened 2016-07-28
Add a protocol version

### LINKED ISSUE #128: Indicate ACME version in /directory (or elsewhere?)
I see nothing in the spec about versioning the protocol.

Should there be information about this in the /directory endpoint, or perhaps a separate endpoint?

bifurcation:
Fixes #128


bifurcation:
Consensus on the list that this is not necessary.

PR #164, opened 2016-07-28
Unparallelize signatures on key-change

### LINKED ISSUE #148: Un-parallelize the signatures on the key-change request
Looking at the current `key-change` mechanism in a little more detail, I have a few concerns with the parallel-signature approach that's there now:

* It's not clear which signature is by the old key and which is by the new key, and the new key doesn't sign over the old key. 
* Both signatures have to include the `"nonce"` parameter, so this request burns two nonces unnecessarily
* The server's JWS checking code now has to switch between validating one signature and two
* Lots of JWS implementations don't support multiple signatures
* If not for this, we could require the flattened JSON serialization
* You need a lot of duplication to deal with the copy/paste attacks (cf. #149).

(On the other hand, I like that both signatures cover the `"url"` parameter; reduces replayability.)

I would propose that we change the signatures from parallel back to serial, as illustrated below.  It would cost a little bit of payload bloat (due to another layer of base64url encoding), but I think the benefits in terms of conceptual simplicity would be worth it.

```
OLD:

{
  "payload": base64url({
    "key": /* New key in JWK form */
  }),
  "signatures": [{
    "protected": base64url({
      "alg": "ES256",
      "jwk": /* old key */,
      "nonce": "pq00v-D1KB0sReG4jFfqVg",
      "url": "https://example.com/acme/key-change"
    }),
    "signature": "XFvVbo9diBlIBvhE...UI62sNT6MZsCJpQo"
  }, {
    "protected": base64url({
      "alg": "ES256",
      "jwk": /* new key */,
      "nonce": "vYjyueEYhMjpVQHe_unw4g",
      "url": "https://example.com/acme/key-change"
    }),
    "signature": "q20gG1f1r9cD6tBM...a48h0CkP11tl5Doo"
  }]
}


NEW:

{
  "protected": base64url({
    "alg": "ES256",
    "jwk": /* old key in JWK form */,
    "nonce": "pq00v-D1KB0sReG4jFfqVg",
    "url": "https://example.com/acme/key-change"
  }),
  "payload": base64({
    "protected": base64url({
      "alg": "ES256",
      "jwk": /* new key in JWK form */,
      "url": "https://example.com/acme/key-change"
    }),
    "payload": base64url({
      "oldKey": /* old key in JWK form */
    }),
    "signature": "q20gG1f1r9cD6tBM...a48h0CkP11tl5Doo"
  }),
  "signature": "XFvVbo9diBlIBvhE...UI62sNT6MZsCJpQo"
}
```

bifurcation:
Fixes #148 

As discussed (too ambiguously) at IETF 96, this refactor uses two nested JWS objects instead of one parallel JWS object.

This preserves the pattern that every ACME message sent by the client is signed by exactly one account key, and saves clients from having to search through multiple signatures in the verification process.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -312,11 +312,11 @@ is readable.  Some fields are omitted for brevity, marked with "...".
 
 ## Equivalence of JWKs
 
-At several points in the protocol, it is necessary for the server to compare one
-key to another in JWK form.  In performing these comparisons, the server SHOULD
-consider two JWKs to be equal if they have the same "kty" value and contain
-identical values in the fields used in the computation of a JWK thumbprint for
-that key type:
+At several points in the protocol, it is necessary for the server to determine
+whether two JWK objects represent the same key.  In performing these checks, the
+server SHOULD consider two JWKs to match if and only if they have the same "kty"
+value and contain identical values in the fields used in the computation of a
+JWK thumbprint for that key type:
 
 * "RSA": "n", "e"
 * "EC": "crv", "x", "y"
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -310,6 +310,17 @@ serialization, with the protected header and payload expressed as
 base64url(content) instead of the actual base64-encoded value, so that the content
 is readable.  Some fields are omitted for brevity, marked with "...".
 
+## Equivalence of JWKs
+
+At several points in the protocol, it is necessary for the server to compare one
+key to another in JWK form.  In performing these comparisons, the server SHOULD
+consider two JWKs to be equal if they have the same "kty" value and contain
+identical values in the fields used in the computation of a JWK thumbprint for
+that key type:
+
+* "RSA": "n", "e"
+* "EC": "crv", "x", "y"
+
 ## Request URI Integrity
 
 It is common in deployment the entity terminating TLS for HTTPS to be different
@@ -991,11 +1002,11 @@ exact string provided in the Location header field in response to the
 new-registration request that created the account.
 
 oldKey (required, JWK):
-: The JWK thumbprint of the original key (i.e., the client's current account
+: The JWK representation of the original key (i.e., the client's current account
 key)
 
 newKey (requrired, JWK):
-: The JWK thumbprint of the new key
+: The JWK representation of the new key
 
 Both of these thumbprints MUST be computed as specified in {{!RFC7638}}, using
 the SHA-256 digest.  The values in the "oldKey" and "newKey" fields MUST be the
@@ -1042,8 +1053,8 @@ Content-Type: application/jose+json
     }),
     "payload": base64url({
       "account": "https://example.com/acme/reg/asdf",
-      "oldKey": "bHcFJ3p0fMo5I6Nu9uvx5Yat04ePLpe0UsA40nNyweg",
-      "newKey": "Nyh_IpZjLIiAu_xm4Xp0Ac5Ckd1Gq1p-UtQjhMADCZI"
+      "oldKey": /* old key */,
+      "newKey": /* new key */
     })
     "signature": "Xe8B94RD30Azj2ea...8BmZIRtcSKPSd8gU"
   }),
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -322,12 +322,19 @@ is readable.  Some fields are omitted for brevity, marked with "...".
 
 At several points in the protocol, it is necessary for the server to determine
 whether two JWK objects represent the same key.  In performing these checks, the
-server SHOULD consider two JWKs to match if and only if they have the same "kty"
-value and contain identical values in the fields used in the computation of a
-JWK thumbprint for that key type:
+server MUST consider two JWKs to match if and only if they have the identical
+values in all fields included in the computation of a JWK thumbprint for that
+key. That is, the keys must have the same "kty" value and contain identical
+values in the fields used in the computation of a JWK thumbprint for that key
+type:
 
 * "RSA": "n", "e"
 * "EC": "crv", "x", "y"
+
+Note that this comparison is equivalent to computing the JWK thumbprints of the
+two keys and comparing thumbprints.  The only difference is that there is no
+requirement for a hash computation (and thus it is independent of the choice of
+hash function) and no risk of hash collision.
 
 ## Request URI Integrity
 
@@ -1001,7 +1008,7 @@ oldKey (required, JWK):
 : The JWK representation of the original key (i.e., the client's current account
 key)
 
-newKey (requrired, JWK):
+newKey (required, JWK):
 : The JWK representation of the new key
 
 Both of these thumbprints MUST be computed as specified in {{!RFC7638}}, using
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -982,7 +982,7 @@ A client may wish to change the public key that is associated with a
 registration in order to recover from a key compromise or proactively mitigate
 the impact of an unnoticed key compromise.
 
-To change the key associate with an account, the client first constructs a
+To change the key associated with an account, the client first constructs a
 key-change object describing the change that it would like the server to make:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -982,65 +982,73 @@ A client may wish to change the public key that is associated with a
 registration in order to recover from a key compromise or proactively mitigate
 the impact of an unnoticed key compromise.
 
-To change the key associated with an account, the client sends a POST request
-containing a key-change object with the following fields:
+To change the key associate with an account, the client first constructs a
```
AGWA: s/associate/associated/

draft-ietf-acme-acme.md
```diff
@@ -982,65 +982,73 @@ A client may wish to change the public key that is associated with a
...
+key-change object describing the change that it would like the server to make:
 
 oldKey (required, JWK):
-: The JWK representation of the original key (i.e., the client's current account
+: The JWK thumbprint of the original key (i.e., the client's current account
 key)
 
 newKey (requrired, JWK):
-: The JWK representation of the new key
+: The JWK thumbprint of the new key
```
ekr: Are you going to commit to the hash function?

draft-ietf-acme-acme.md
```diff
@@ -982,65 +982,73 @@ A client may wish to change the public key that is associated with a
...
 
-The JWS of this POST must have two signatures: one signature from the existing
-key on the account, and one signature from the new key that the client proposes
-to use. This demonstrates that the client actually has control of the
-private key corresponding to the new public key. The protected header must
-contain a JWK field containing the current account key.
+The client then encapsulates the key-change object in a JWS, signed with the
+client's current account key (i.e., the key matching the "oldKey" value).  This
+JWS is NOT REQUIRED to have the "nonce" and "url" fields, despite the general
+requirement that these fields be present.
```
ekr: Why not the url field?

draft-ietf-acme-acme.md
```diff
@@ -982,65 +982,73 @@ A client may wish to change the public key that is associated with a
...
+This inner JWS then become the payload of the JWS that the client sends to the
+server.  The outer JWS is signed with the key pair that the the client wishes to
+have the server use as its account key (i.e., the key pair matching the "newKey"
+value).  The client sends this outer JWS in a POST request to the server's
+"key-change" resource.
+
+This transaction has signatures from both the old and new keys so that the
+server can verify that the holders of the two keys both agree to the change.
+The signatures are nested to preserve the property that all signatures on POST
+messages are signed by exactly one key.
```
ekr: Can? MUST verify, right?

draft-ietf-acme-acme.md
```diff
@@ -982,65 +982,73 @@ A client may wish to change the public key that is associated with a
...
+The client then encapsulates the key-change object in a JWS, signed with the
+client's current account key (i.e., the key matching the "oldKey" value).  This
+JWS is NOT REQUIRED to have the "nonce" and "url" fields, despite the general
+requirement that these fields be present.
+
+This inner JWS then become the payload of the JWS that the client sends to the
+server.  The outer JWS is signed with the key pair that the the client wishes to
+have the server use as its account key (i.e., the key pair matching the "newKey"
+value).  The client sends this outer JWS in a POST request to the server's
+"key-change" resource.
```
ekr: Why are you signing the outer one with the new key? It seems more natural to sign the inner one with the new key because then you can use the general input procesing logic. Is there a cryptographic reason?

draft-ietf-acme-acme.md
```diff
@@ -982,65 +982,73 @@ A client may wish to change the public key that is associated with a
...
+  "payload": base64url({
     "protected": base64url({
       "alg": "ES256",
       "jwk": /* new key */,
-      "nonce": "vYjyueEYhMjpVQHe_unw4g",
-      "url": "https://example.com/acme/key-change"
     }),
-    "signature": "q20gG1f1r9cD6tBM...a48h0CkP11tl5Doo"
-  }]
+    "payload": base64url({
```
ekr: Is something wrong here? The previous version had one jwk labeled old key and one labelled new key but now...

draft-ietf-acme-acme.md
bifurcation: Well, yes, but that's encoded below.

draft-ietf-acme-acme.md
bifurcation: Not a cryptographic reason, more forward-compatibility.  Namely, if you had a way to provision a MAC key as a recovery key, then you could replace the outer `newKey` signature with a MAC by the recovery key, and voilà, a recovery mechanism.  But since we're not doing recovery right now, maybe it's less important.

draft-ietf-acme-acme.md
bifurcation: Yeah, copy/paste error.

draft-ietf-acme-acme.md
bifurcation: Good point.  I think we probably should, as we do for key authorizations below (vs. specifying one in the object).  We can update later by defining a new `key-change` endpoint.  

We also need to specify that the thumbprint is base64url-encoded, since that's not in RFC 7638 either.  Maybe the thing to do is just to say somewhere at the beginning that when we say "thumbprint" we mean "... with SHA-256, base64url-encoded".

draft-ietf-acme-acme.md
bifurcation: It seems like design-wise, there are two real alternatives:

1. Same requirements as the outer JWS
2. No ACME-specific requirements

I had in mind (2) here because it seemed like a hassle to burn two nonces at once.  Also, the benefit of the "url" field here seems pretty small, since (at least in the context of ACME), this object can only be replayed by the holder of the old account key, and doesn't isn't useful for anything other than rolling a key.

But hey, rigidity is good and nonces are plentiful, so why not (1).  You also get the benefit of being able to always require `nonce` and `url` in your `sign()` method.  The main cost is going to be the need for a more nonce-priming HEAD requests, but since this is an infrequent operation, that doesn't seem tragic.

AGWA: Good catch, @ekr.  It's actually very important to include the url in the inner JWS.  Otherwise, the holder of the new key has the ability to take over accounts using the old key on any ACME server, by replaying the inner JWS.  If the same person controls both keys, that's not a big deal, but if you're transferring your ACME account to a different person, it's quite important to prevent this.

It's also important, albeit less so, to include the nonce in the inner JWS.  Otherwise if you create a new account using the old key, the holder of the new key can take it over.  That's only a problem if you transfer your account to a different person and then reuse your old key for a new account.

draft-ietf-acme-acme.md
```diff
@@ -318,6 +318,17 @@ serialization, with the protected header and payload expressed as
 base64url(content) instead of the actual base64-encoded value, so that the content
 is readable.  Some fields are omitted for brevity, marked with "...".
 
+## Equivalence of JWKs
+
+At several points in the protocol, it is necessary for the server to determine
+whether two JWK objects represent the same key.  In performing these checks, the
+server SHOULD consider two JWKs to match if and only if they have the same "kty"
```
ekr: Why is this a SHOULD and not a MUST

draft-ietf-acme-acme.md
```diff
@@ -318,6 +318,17 @@ serialization, with the protected header and payload expressed as
...
+value and contain identical values in the fields used in the computation of a
+JWK thumbprint for that key type:
+
+* "RSA": "n", "e"
+* "EC": "crv", "x", "y"
+
```
ekr: Why isn't this just telling people "compute the fingerprint"

draft-ietf-acme-acme.md
```diff
@@ -988,55 +1004,77 @@ key)
 newKey (requrired, JWK):
```
ekr: requrired is not a word

draft-ietf-acme-acme.md
bifurcation: JSHA objected to the additional fingerprint computations.  I will rephrase, though, to make clear that this is equivalent to comparing fingerprints, modulo hash collisions.

PR #165, opened 2016-08-05
Re-add new-authz as pre-authorization

bifurcation:
As discussed at IETF 96, even with the new-app flow, it is still useful to be able to request authorization outside of an issuance transaction.  This PR re-adds the new-authz transaction.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -509,23 +509,23 @@ indicate HTTP link relations
                                directory
                                    |
                                    |
-       ----------------------------------------------------
-       |                  |                               |
-       |                  |                               |
-       V                  V                               V
-    new-reg            new-app                       revoke-cert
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1272,13 +1368,16 @@ specific application by setting the "scope" field of the authorization to the
...
-implicitly tied to the account key used to sign the new-application request.
+applications or authorization requests submitted by an account key holder; their
+URLs are provided to the client in the responses to these requests.  The
+authorization object is implicitly tied to the account key used to sign the
+request.
 
 When a client receives an application from the server with an "authorization"
 requirement, it downloads the authorization resource by sending a GET request to
-the indicated URL.
+the indicated URL.  If the client initiates authorization using a request to the
```
parche-model: Www.parche-model.com

PR #167, opened 2016-08-09
Simplify terms-of-service flow.

jsha:
This mitigates confusion about how clients should respond when terms of service change.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -920,11 +920,11 @@ registration in a Location header field.  This allows a client that has an
 account key but not the corresponding registration URI to recover the
 registration URI.
 
-If the server provides a terms-of-service URL in the directory, the client MUST
-indicate its agreement to the terms at that URL by including the
-"terms-of-service": "agreed" field in the new-registration body. Servers SHOULD
-reject registration requests that do not meet their requirements for terms of
-service.
+If the server wishes to present the client with terms under which the ACME
+service is to be used, it MUST indicate the URI where such terms can be accessed
+in a Link header with link relation "terms-of-service".  As noted above, the
+client may indicate its agreement when creating registraion by including the
+"terms-of-service": "agreed" field.
 
 ~~~~~~~~~~
 HTTP/1.1 201 Created
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -173,8 +173,7 @@ The first phase of ACME is for the client to register with the ACME server.  The
 client generates an asymmetric key pair and associates this key pair with a set
 of contact information by signing the contact information.  The server
 acknowledges the registration by replying with a registration object echoing the
-client's input.  The server can also provide terms of service at this stage,
-which the client can present to a human user.
+client's input.
 
 ~~~~~~~~~~
       Client                                                  Server
@@ -183,7 +182,6 @@ which the client can present to a human user.
       Signature                     ------->
 
                                     <-------            Registration
-                                                    Terms of Service
 ~~~~~~~~~~
 
 
@@ -886,6 +884,7 @@ Content-Type: application/jose+json
     "url": "https://example.com/acme/new-reg"
   })
   "payload": base64url({
+    "terms-of-service": "agreed",
     "contact": [
       "mailto:cert-admin@example.com",
       "tel:+12025551212"
@@ -913,19 +912,16 @@ registration in a Location header field.  This allows a client that has an
 account key but not the corresponding registration URI to recover the
 registration URI.
 
-If the server wishes to present the client with terms under which the ACME
-service is to be used, it MUST indicate the URI where such terms can be accessed
-in a Link header with link relation "terms-of-service".  As noted above, the
-client may indicate its agreement with these terms by updating its registration
-to include the "agreement" field, with the terms URI as its value.  When these
-terms change in a way that requires an agreement update, the server MUST
-use a different URI in the Link header.
+If the server provides a terms-of-service URL in the directory, the client MUST
+indicate its operator's agreement to the terms at that URL by including the
+"terms-of-service": "agreed" field in the new-registration body. Servers SHOULD
+reject registration requests that do not meet their requirements for terms of
+service.
 
 ~~~~~~~~~~
 HTTP/1.1 201 Created
 Content-Type: application/json
 Location: https://example.com/acme/reg/asdf
-Link: <https://example.com/acme/terms>;rel="terms-of-service"
 Link: <https://example.com/acme/some-directory>;rel="directory"
 
 {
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -924,16 +923,13 @@ registration URI.
 If the server wishes to present the client with terms under which the ACME
 service is to be used, it MUST indicate the URI where such terms can be accessed
 in a Link header with link relation "terms-of-service".  As noted above, the
-client may indicate its agreement with these terms by updating its registration
-to include the "agreement" field, with the terms URI as its value.  When these
-terms change in a way that requires an agreement update, the server MUST
-use a different URI in the Link header.
+client may indicate its agreement when creating registraion by including the
```
konklone: Typo: `registraion` -> `registration`.

draft-ietf-acme-acme.md
```diff
@@ -894,6 +892,7 @@ Content-Type: application/jose+json
     "url": "https://example.com/acme/new-reg"
   })
   "payload": base64url({
+    "terms-of-service": "agreed",
```
bifurcation: Let's do a boolean instead of a fixed string, e.g.:

```
  "terms-of-service-agreed": true,
```

bifurcation: You're also going to need to integrate this change in a couple of other places above.  Instances I've noticed:

* "agreement (optional, string):" et seq.
* "The body of the request is a stub registration object containing only the "contact" field"

jsha:
Updated based on review. Took out the echoing of the field on subsequent requests, because it makes things simpler and clearer.

draft-ietf-acme-acme.md
```diff
@@ -662,11 +660,6 @@ contact (optional, array of string):
 related to this authorization. For example, the server may wish to notify the
 client about server-initiated revocation.
 
-agreement (optional, string):
-: A URI referring to a subscriber agreement or terms of service provided by the
-server (see below).  Including this field indicates the client's agreement with
-the referenced terms.
-
```
bifurcation: You need to add the "terms-of-serivce-agreed" field.

draft-ietf-acme-acme.md
```diff
@@ -955,18 +948,17 @@ key but not the corresponding registration URI to recover the registration URI.
...
-in a Link header with link relation "terms-of-service".  As noted above, the
-client may indicate its agreement with these terms by updating its registration
-to include the "agreement" field, with the terms URI as its value.  When these
-terms change in a way that requires an agreement update, the server MUST
-use a different URI in the Link header.
+in the "terms-of-service" subfield of the "meta" field in the directory object,
+and the server MUST reject new-registration requests that do not contain
+"terms-of-service-agreed": true. Note that the "terms-of-service-agreed" field
+is not echoed in the resulting registration object, because it is a one-time
+field.
```
bifurcation: I'm not loving "because it is a one-time field".  Basically, you're saying there's no need to reflect this on the registration object, because it's implicitly true given that the server rejects new-reg when it's false.

I don't think that's actually true, though, if the server can make breaking changes to ToS and clients need to re-agree.  In that case, there's no way to tell from the registration object whether you've already re-agreed or not; you just have to try something and get the error.  And in any case, explicit state > implicit state.

Net/net, I would prefer if we could just have it be a field on the object; the server's going to have to manage it either way.

draft-ietf-acme-acme.md
```diff
@@ -2354,7 +2346,6 @@ certificate applicant controls the identifier he claims.  Before issuing a
 certificate, however, there are many other checks that a CA might need to
 perform, for example:
 
-* Has the client agreed to a subscriber agreement?
```
bifurcation: Why would we delete this?  CAs do need to check this (e.g., to comply with the BRs), even if the code for it isn't in the `issue()` method in their ACME server.

draft-ietf-acme-acme.md
cpu: I'm +1 on echoing it back - I don't think it would upset clients and is less special-casing.

bifurcation: This sentence is now false.

draft-ietf-acme-acme.md
```diff
@@ -662,10 +660,10 @@ contact (optional, array of string):
...
 client about server-initiated revocation.
 
-agreement (optional, string):
-: A URI referring to a subscriber agreement or terms of service provided by the
-server (see below).  Including this field indicates the client's agreement with
-the referenced terms.
+terms-of-service-agreed (optional, boolean):
+: Including this field in a new-registration request, with a value of true,
+indicates the client's agreement with the terms of service. This field is not
+updateable by the client.
```
bifurcation: This isn't really true, because clients might need to re-agree.  (And we don't want them to have to re-register.)  How about the following?

```
Once set to "true" by the client, the value of this field cannot be changed by the client.  If the server requires the client to refresh its agreement to the terms, it does so by clearing this field and accepting a registration update to reset it to "true".
```

draft-ietf-acme-acme.md
```diff
@@ -955,18 +954,17 @@ key but not the corresponding registration URI to recover the registration URI.
 
 If the server wishes to present the client with terms under which the ACME
 service is to be used, it MUST indicate the URI where such terms can be accessed
-in a Link header with link relation "terms-of-service".  As noted above, the
-client may indicate its agreement with these terms by updating its registration
-to include the "agreement" field, with the terms URI as its value.  When these
-terms change in a way that requires an agreement update, the server MUST
-use a different URI in the Link header.
+in the "terms-of-service" subfield of the "meta" field in the directory object,
+and the server MUST reject new-registration requests that do not contain
```
bifurcation: This requirement is not appropriate.  It's a matter of server policy when it requires agreement. The server can respond with an `agreementRequired` error code whenever it wants to.

jsha: Again, I'm trying to simplify the protocol here. There is zero reason for the protocol to support a server that offers a ToS URL but treats them as optional for registering an account. This only creates confusion.

draft-ietf-acme-acme.md
```diff
@@ -980,18 +985,15 @@ key but not the corresponding registration URI to recover the registration URI.
 
 If the server wishes to present the client with terms under which the ACME
 service is to be used, it MUST indicate the URI where such terms can be accessed
-in a Link header with link relation "terms-of-service".  As noted above, the
-client may indicate its agreement with these terms by updating its registration
-to include the "agreement" field, with the terms URI as its value.  When these
-terms change in a way that requires an agreement update, the server MUST
-use a different URI in the Link header.
+in the "terms-of-service" subfield of the "meta" field in the directory object,
+and the server MAY reject new-registration requests that do not contain
```
bifurcation: I would s/MAY/SHOULD/

draft-ietf-acme-acme.md
bifurcation: Summarizing offline comment: Even though this seems inconsistent, we're going to only support initial agreement in-band, not re-agreement.  The rare CA that needs re-agreement can use the mechanism we're in #182.

PR #171, opened 2016-08-17
Remove combinations array.

jsha:
This is a fairly complicated part of the protocol, and not used in practice. For
instance, in Let's Encrypt's implementation, there are always three challenges,
any one of which may be fulfilled by the client.

After this change, all challenges are considered to be combined with an "OR."
That is, any challenge within an authorization may be completed to make the
authorization valid.

Authorizations within the new-application object are considered to be combined
with an "AND." That is, all of them must become valid before the certificate
will be issued. The combination of the two means that we have similar
expressiveness as before, even without the combinations array.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -826,19 +826,13 @@ applications until the authorization expires. [[ Open issue: More flexible
 scoping? ]]
 
 challenges (required, array):
-: The challenges that the client needs to fulfill
+: The challenges that the client may to fulfill
 in order to prove possession of the identifier (for pending authorizations).
 For final authorizations, the challenges that were used.  Each array entry is a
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -826,19 +826,13 @@ applications until the authorization expires. [[ Open issue: More flexible
 scoping? ]]
 
 challenges (required, array):
-: The challenges that the client needs to fulfill
+: The challenges that the client may to fulfill
```
bifurcation: This doesn't parse as-is.  Suggest "The challenges that the client can fulfill", with clarification that any one of them is sufficient below.

jsha: Updated.

Issue #170, opened 2016-08-11
add a special token parameter in registration


startcomca:
StartCom plan to use ACME protocol for StartEncrypt, we need to identify the client's validation level, so the subscriber administration can generate a special token in the StartSSL.com account that send this token to the email address used in the registration.
At the registration, user need to enter email and this token with the certificate to let the CA system know this customer's validation level. 
After the CA system receive the email, the token and signing certificate, CA system know what type of certificate we can issue to this client; if this client account is class 4 validated, then the client can get EV SSL certificate, not DV SSL.
please add this a parameter to the ACME protocol, thanks.

rolandshoemaker:
This request should be sent to the ACME mailing list (acme@ietf.org) for further discussion, it would also be very useful to know what your EV (or OV) validation flow would look like using ACME as most likely the current certificate application object would need to be changed to identify those requirements to the server.

bifurcation:
This was fixed by #212

PR #172, opened 2016-08-18
Add an external secret field to registration.

jsha:
This is one option to try and solve
https://github.com/ietf-wg-acme/acme/issues/170.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -634,6 +634,11 @@ contact (optional, array of string):
 related to this authorization. For example, the server may wish to notify the
 client about server-initiated revocation.
 
+external_secret (optional, string):
+: An array of URIs that the server can use to contact the client for issues
+related to this authorization. For example, the server may wish to notify the
+client about server-initiated revocation.
+
 agreement (optional, string):
 : A URI referring to a subscriber agreement or terms of service provided by the
 server (see below).  Including this field indicates the client's agreement with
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -894,6 +904,7 @@ Content-Type: application/jose+json
     "url": "https://example.com/acme/new-reg"
   })
   "payload": base64url({
+    "external_secret": "XYZ123",
```
cpu: On L637 `external_secret` is described as a "array of URIs". Should the example be in `[ ]` brackets? I.e.: `"external_secret": [ "XYZ123" ],` ?

cpu: I might have confused myself. Perhaps the `external_secret` description on L637 copied too closely the `contact` field?

Based on the discussion from #170 it seems like this field isn't intended to be used to "contact the client" at all and is later described (L935+) more accurately as an existing authenticator that the server can use to associate the client to an external system.

jsha: This was my mistake. I copied the `contact` field description but failed to edit it appropriately. Fixed now.

cpu: @jsha thanks!

draft-ietf-acme-acme.md
```diff
@@ -635,9 +635,8 @@ related to this authorization. For example, the server may wish to notify the
 client about server-initiated revocation.
 
 external_secret (optional, string):
-: An array of URIs that the server can use to contact the client for issues
-related to this authorization. For example, the server may wish to notify the
-client about server-initiated revocation.
+: A secret value that the server can use to authenticate this registration as
+belonging the same entity as an already-existing system.
```
cpu: There's an additional typo here:
> belonging the same entity

`s/belonging the/belonging to the/g`

PR #182, opened 2016-08-26
Clarify flows around agreement to terms

bifurcation:
Right now, it's not clear when the client needs to agree to the server's terms, and what the server should do if it needs the client to re-agree.  This PR addresses both with a dedicated error code and some related behaviors.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1020,9 +1020,9 @@ That is, it should send a JWS whose payload is trivial ({}).
 
 ### Agreement to Terms of Service
 
-As described above, a client can indicate its agreement with a specific set of
-terms (referenced by URL) by indicating that URL in the "agreement" field of its
-registration object.
+As described above, a client can indicate its agreement with the CA's terms of
+service by setting the "terms-of-service-agreed" field in its registration
+object to "true".
 
 If a client has not agreed to the server's terms of service and the server is
 unwilling to process a request without agreement, then it MUST return an error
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -972,6 +973,31 @@ information about its account (e.g., to examine the "contact" or "certificates"
...
+terms-of-service URL.
+
+If a client receives an "agreementRequired" error and is willing to agree to the
+terms of service referenced in the Link header, then it SHOULD update its
+registration with the new agreement URL and retry the original request.
+
+The "agreementRequired" error allows CAs to clearly express when agreement to
+terms is required, both for new registrations that have not agreed to any terms,
+and for existing registrations that may have agreed to an older version of the
+terms.  If a CA makes a change to its terms of service or terms-of-service URL
```
cpu: There's some extra whitespace before "If a CA makes"

draft-ietf-acme-acme.md
```diff
@@ -1051,17 +1051,30 @@ response with status code 403 (Forbidden) and type
...
+~~~~~
+HTTP/1.1 403 Forbidden
+Replay-Nonce: IXVHDyxIRGcTE0VSblhPzw
+Content-Type: application/problem+json
+Content-Language: en
+
+{
+  "type": "urn:ietf:params:acme:error:agreementRequired"
+  "detail": "Terms of service have changed"
+  "action": "http://example.com/agreement/?token=W8Ih3PswD-8"
```
jsha: Rather than defining a new field, let's just use the `instance` field already defined in [RFC7807](https://tools.ietf.org/html/rfc7807#section-3.1)

draft-ietf-acme-acme.md
```diff
@@ -1051,17 +1051,30 @@ response with status code 403 (Forbidden) and type
...
+}
+~~~~~
+
+If an "action" field is provided, then the client indicates its agreement by
```
jsha: This is an example of "action happens just by clicking," which it not good. Instead, the URL should contain human-readable instructions about how to read the new ToS and agree.

draft-ietf-acme-acme.md
```diff
@@ -1051,17 +1051,30 @@ response with status code 403 (Forbidden) and type
...
+visiting the indicated URL.  Otherwise, the client update sits registration to
```
jsha: In https://github.com/ietf-wg-acme/acme/pull/167 we're saying that "terms-of-service-agreed" is only changeable by the server, so the rest of this paragraph should go.

bifurcation: In that case, we need to change the "SHOULD reject new-reg without agreement" to a MUST, in order to avoid having clients registered who can't now agree.

draft-ietf-acme-acme.md
bifurcation: Not what I meant.  Will clarify.

draft-ietf-acme-acme.md
```diff
@@ -1059,6 +1060,37 @@ information about its account (e.g., to examine the "contact" or "certificates"
...
 That is, it should send a JWS whose payload is trivial ({}).
 
+### Changes of Terms of Service
+
+As described above, a client can indicate its agreement with the CA's terms of
+service by setting the "terms-of-service-agreed" field in its registration
+object to "true".
+
+If the server has changed its terms of service since a client initially agreed,
+and the server is unwilling to process a request without agreement to the new
```
jsha: Let's make this "without explicit agreement."

PR #183, opened 2016-08-26
Add 'revoked' status to registrations and use standard language

rolandshoemaker:


bifurcation:
what's the difference between "deactivated" and "revoked"?

jsha:
Deactivated is done by the user; revoked is done by an administrator.

bifurcation:
OK, so we need to either say that or use clearer terms :)  I talked with @rolandshoemaker offline, and I think he's going to propose something.

PR #184, opened 2016-08-26
Describe how IDNs should be encoded

rolandshoemaker:


Jay838:
I don't see any technical reason why characters should be restricted at all. It is not up to ACME to decide what the TLD/gTLD providers want to allow for encoding.

The text should rather say something like:

Domain names must be represented in the same character encoding as the domain is registered with the registrar.
For IDN domains, the domain must be encoded in IDN/punycode.
For UTF-8 domains, the domain must be encoded in UTF-8 and so on.
The same applies to UTF-16, ISO8859-x and any other character encoding.

rolandshoemaker:
[RFC5280 section 7.2](https://tools.ietf.org/html/rfc5280#section-7.2) requires IDNs be encoded in this way in order to be stored in the X509 certificate dNSNames field. Compliance with this RFC is one of the requirements of the CABForum [Baseline Requirements](https://cabforum.org/baseline-requirements-documents/) which CAs are expected to adhere to.

We already require domain names transmitted via ACME use the ASCII encoding, this just formalizes how IDNs should be converted.

draft-ietf-acme-acme.md
```diff
@@ -838,9 +838,12 @@ one of these challenges, and a server should consider any one of the challenges
 sufficient to make the authorization valid.
 
 The only type of identifier defined by this specification is a fully-qualified
-domain name (type: "dns").  The value of the identifier MUST be the ASCII
-representation of the domain name.  Wildcard domain names (with "*" as the first
-label) MUST NOT be included in authorization requests.
+domain name (type: "dns"). The value of the identifier MUST be the ASCII
+representation of the domain name. If a domain name contains Unicode characters
+it should be encoded using the rules defined in {{!RFC3492}}. Servers MUST verify
```
bifurcation: "should" -> "MUST"

draft-ietf-acme-acme.md
```diff
@@ -838,9 +838,12 @@ one of these challenges, and a server should consider any one of the challenges
...
+any identifier values that begin with the ASCII Compatible Encoding prefix as
```
bifurcation: I would go ahead and note the prefix here to save people some reading:

```
ASCII Compatible Encoding prefix "xn--"...
```

PR #186, opened 2016-09-06
Remove extraneous statuses

rolandshoemaker:
Remove unknown status from applications and authorizations and processing status from applications

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -672,8 +672,8 @@ server's requirements, and any certificates that have resulted from this
 application.
 
 status (required, string):
-: The status of this authorization.  Possible values are: "unknown", "pending",
-"processing", "valid", and "invalid".
+: The status of this authorization.  Possible values are: "pending", "valid",
+and "invalid".
 
 expires (optional, string):
 : The timestamp after which the server will consider this application invalid,
@@ -812,9 +812,9 @@ identifier (required, dictionary of string):
   : The identifier itself.
 
 status (required, string):
-: The status of this authorization.  Possible values are: "unknown", "pending",
-"processing", "valid", "invalid" and "revoked".  If this field is missing, then
-the default value is "pending".
+: The status of this authorization.  Possible values are: "pending", "processing",
+"valid", "invalid" and "revoked".  If this field is missing, then the default
+value is "pending".
 
 expires (optional, string):
 : The timestamp after which the server will consider this authorization invalid,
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -812,9 +812,9 @@ identifier (required, dictionary of string):
   : The identifier itself.
 
 status (required, string):
-: The status of this authorization.  Possible values are: "unknown", "pending",
-"processing", "valid", "invalid" and "revoked".  If this field is missing, then
-the default value is "pending".
+: The status of this authorization.  Possible values are: "pending", "processing",
```
bifurcation: I think we can probably remove "processing" here too.  I originally had this idea that we would want to express whether the thing was waiting on client action or server action, but I'm not seeing any cases here where there would be ambiguity.

rolandshoemaker: On 09/08/2016 10:12 AM, bifurcation wrote:
> In draft-ietf-acme-acme.md
> <https://github.com/ietf-wg-acme/acme/pull/186#discussion_r78048809>:
> 
>> @@ -812,9 +812,9 @@ identifier (required, dictionary of string):
>>    : The identifier itself.
>>  
>>  status (required, string):
>> -: The status of this authorization.  Possible values are: "unknown", "pending",
>> -"processing", "valid", "invalid" and "revoked".  If this field is missing, then
>> -the default value is "pending".
>> +: The status of this authorization.  Possible values are: "pending", "processing",
> 
> I think we can probably remove "processing" here too. I originally had
> this idea that we would want to express whether the thing was waiting on
> client action or server action, but I'm not seeing any cases here where
> there would be ambiguity.
> 

I think if a server were to implement validation retries for some period
it might make sense to have a 'processing' status that would be used
during that time. Not sure if that is really that useful though...

jsha: I'm in favor of keeping "processing." Boulder doesn't currently implement that status but it should.

bifurcation: I can live with that.

draft-ietf-acme-acme.md
```diff
@@ -672,8 +672,8 @@ server's requirements, and any certificates that have resulted from this
 application.
 
 status (required, string):
-: The status of this authorization.  Possible values are: "unknown", "pending",
-"processing", "valid", and "invalid".
+: The status of this authorization.  Possible values are: "pending", "valid",
+and "invalid".
```
bifurcation: s/authorization/application/

PR #189, opened 2016-09-15
Fix signing order for key-change endpoint.

jsha:
Signing with the current account key allows servers to use the same message
transport authentication that they use for other requests, simplifying the
verification.

Also fix two spots where we still described newKey/oldKey as thumbprints.


bifurcation:
LGTM.  I had only done that signing order to be forward-compatible with some imagined future recovery scheme, but if we decide to do that someday, that patch can differ in signature order as well as whatever else.

PR #191, opened 2016-09-20
Standardize on "proactive" certificate issuance.

cpu:
Based on the acme mailing list discussion[0] on "proactive" vs "on-finalization" issuance it seems there is more support for the proactive approach.

Speaking on behalf of Let's Encrypt and the Boulder developers we're willing to compromise and support proactive issuance as the sole issuance method in order to simplify the protocol & implementations.

This PR changes the language around certificate issuance to reflect that the server MUST issue the certificate proactively once all the required challenges for an application have been fulfilled. The open
issue corresponding to making this decision is removed now that there is one supported option.

In an effort to close out open questions around applications this commit also removes the open issue around multiple certificates per application. If there is a demand for automating repeat issuance there should be a new thread started with a concrete proposal.

0: https://mailarchive.ietf.org/arch/msg/acme/0lVmGl8e-rmSH0x7ycDW5dj6GAY

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -187,11 +187,12 @@ which the client can present to a human user.
 ~~~~~~~~~~
 
 
-Once the client is registered, there are two major steps it needs to take to
+Once the client is registered, there are three major steps it needs to take to
 get a certificate:
 
 1. Apply for a certificate to be issued
 2. Fulfill the server's requirements for issuance
+3. Await issuance and download the issued certificate
 
 The client's application for a certificate describes the desired certificate
 using a PKCS#10 Certificate Signing Request (CSR) plus a few additional fields
@@ -209,8 +210,8 @@ The server then validates the challenges to check that the client has
 accomplished the challenge.
 
 Once the validation process is complete and the server is satisfied that the
-client has met its requirements, the server MUST proactively issue the
-requested certificate and make it available to the client.
+client has met its requirements, the server will issue the requested certificate
+and make it available to the client.
 
 ~~~~~~~~~~
       Application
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -187,12 +187,11 @@ which the client can present to a human user.
 ~~~~~~~~~~
 
 
-Once the client is registered, there are three major steps it needs to take to
+Once the client is registered, there are two major steps it needs to take to
 get a certificate:
 
 1. Apply for a certificate to be issued
 2. Fulfill the server's requirements for issuance
-3. Finalize the application and request issuance
```
bifurcation: Maybe change to "Await issuance and download the issued certificate" instead of deleting

draft-ietf-acme-acme.md
```diff
@@ -210,10 +209,8 @@ The server then validates the challenges to check that the client has
 accomplished the challenge.
 
 Once the validation process is complete and the server is satisfied that the
-client has met its requirements, the server can either proactively issue the
-requested certificate or wait for the client to request that the application be
-"finalized", at which point the certificate will be issued and provided to the
-client.
+client has met its requirements, the server MUST proactively issue the
```
bifurcation: Don't think the word "proactively" is needed.  It was useful in the discussion, but not so much now.

Let's drop the "MUST".  It's covered below; this is just an overview.

cpu:
@bifurcation apologies! I missed your comments until today. Both are resolved in https://github.com/ietf-wg-acme/acme/pull/191/commits/1e090367846cfcbcac24648ba61415f95b8fb396 Thank you for the feedback.

cpu:
Oh! They were fresh and I misread the dates. Not such a bad turn around after all :-)

bifurcation:
@cpu Fantastic, thanks.  Merging.

PR #192, opened 2016-09-21
Updates Section 6.1.2 - "Registration Objects" for Applications.

cpu:
In [Section 6.1.2 "Registration Objects"](https://github.com/ietf-wg-acme/acme/blob/83e2f7985f5921a4ac30b7446dbb9c1d347d69ec/draft-ietf-acme-acme.md#registration-objects) the "applications" field of the registration resource is described in great detail. This is confusing for two reasons:

1. both the text and the example registration object are still using the "authorization" term that applications replace.
2. the text is written as if the applications should be directly embedded in the registration object returned.

This PR addresses both issues by describing the "applications" field as an array of application URIs that the client can fetch. The example "registration" object is updated accordingly.

The note indicating a server SHOULD include pending applications and SHOULD NOT include invalid applications remains as-is. The note indicating a server MAY return an incomplete list is removed - it is more complicated and with only URIs being returned, likely an unneeded optimization.

Finally, in [Section 6.3 "Registration"](https://github.com/ietf-wg-acme/acme/blob/83e2f7985f5921a4ac30b7446dbb9c1d347d69ec/draft-ietf-acme-acme.md#registration) there are three fields that must be ignored by the server in a registration body: key, authorizations and certificates. This should be "key" and "applications" and is updated accordingly.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -670,13 +670,9 @@ server (see below).  Including this field indicates the client's agreement with
 the referenced terms.
 
 applications (required, string):
-: A URI from which a list of authorizations submitted by this account can be
-fetched via a GET request.  The result of the GET request MUST be a JSON object
-whose "applications" field is an array of strings, where each string is the URI
-of an authorization belonging to this registration.  The server SHOULD include
-pending applications, and SHOULD NOT include applications that are invalid. The
-server MAY return an incomplete list, along with a Link header with link
-relation "next" indicating a URL to retrieve further entries.
+: An array of URIs from which application objects submitted by this account can
+be fetched via a GET request. The server SHOULD include pending applications,
+and SHOULD NOT include applications that are invalid.
 
 ~~~~~~~~~~
 {
@@ -685,8 +681,10 @@ relation "next" indicating a URL to retrieve further entries.
     "tel:+12025551212"
   ],
   "agreement": "https://example.com/acme/terms",
-  "authorizations": "https://example.com/acme/reg/1/authz",
-  "certificates": "https://example.com/acme/reg/1/cert"
+  "applications": [
+    "https://example.com/acme/reg/1/apps/1",
+    "https://example.com/acme/reg/1/apps/2"
+  ]
 }
 ~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -768,6 +762,8 @@ syntax like "confirm": true ]]
 support reissuance / renewal in a straightforward way, especially if the CSR /
 notBefore / notAfter could be updated. ]]
 
+#### Application Requirements
+
```
bifurcation: Note: These changes are unrelated to the topic of the PR.  I think they're OK, though.

draft-ietf-acme-acme.md
```diff
@@ -818,6 +814,30 @@ url (required, string):
 To fulfill this requirement, the ACME client should direct the user to the
 indicated web page.
 
+#### Applications List
```
bifurcation: I agree that this is better as its own subsection, but it would go better under the "Registration Objects" section, since ultimately it's about how registrations refer to applications.

draft-ietf-acme-acme.md
```diff
@@ -818,6 +814,30 @@ url (required, string):
...
+~~~~~~~~~~
+HTTP/1.1 200 OK
+Content-Type: application/json
+Link: href="/acme/reg/1/apps?cursor=2", rel="next"
+
+{
+  "applications": [
+    "https://example.com/acme/reg/1/apps/1",
+    "https://example.com/acme/reg/1/apps/2"
+  ]
```
bifurcation: For a little more verisimilitude, you might up the numbers a little bit :)

```
HTTP/1.1 200 OK
Content-Type: application/json
Link: href="/acme/reg/1/apps?cursor=51", rel="next"

{
  "applications": [
    "https://example.com/acme/reg/1/apps/1",
    "https://example.com/acme/reg/1/apps/2",
    /* 47 more URLs */
    "https://example.com/acme/reg/1/apps/50"
  ]
}
```

cpu: Good idea ("verisimilitude" is an excellent word!!!)

cpu:
@bifurcation feedback addressed, and conflicts w/ master resolved. Thanks for the :eyes: !

PR #193, opened 2016-09-23
Specify account by kid (reg URL) rather than key.

jsha:
Previously, each authenticated POST included a copy of its public key in the JWS
protected header. Servers were required to canonicalize this key in some way and
look it up in a database, then verify the payload. However, servers also had the
option of verifying the payload, and *then* looking up the key in the database.
This opened up a category of potential bugs based on truncated database fields
or hash collisions, where an attacker could craft a key pair the would
potentially be matched to someone else's account.

Incidental benefits: Switching to kid makes payloads somewhat smaller, and
removes the need to define key equality in the ACME spec.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -308,7 +308,7 @@ JWS objects sent in ACME requests MUST meet the following additional criteria:
   * "nonce" (defined below)
   * "url" (defined below)
 
-For new-reg requests and revoke-cert-by-key requests, there MUST be a
+For new-reg requests, there MUST be a
 "jwk" field containing the key being registered, and there MUST NOT be a
 "kid" field.
 
@@ -519,15 +519,14 @@ ACME is structured as a REST application with a few types of resources:
 * A "new-registration" resource
 * A "new-application" resource
 * A "revoke-certificate" resource
-* A "revoke-certificate-by-key" resource
 * A "key-change" resource
 
 The server MUST provide "directory" and "new-nonce" resources.
 
-For the singular resources above ("directory", "new-nonce", "new-registration",
-"new-application", "revoke-certificate", "revoke-certificate-by-key", and
-"key-change") the resource may be
-addressed by multiple URIs, but all must provide equivalent functionality.
+For the singular resources above ("directory", "new-nonce",
+"new-registration", "new-application", "revoke-certificate", and
+"key-change") the resource may be addressed by multiple URIs, but all
+must provide equivalent functionality.
 
 ACME uses different URIs for different management functions. Each function is
 listed in a directory along with its corresponding URI, so clients only need to
@@ -647,7 +646,6 @@ Content-Type: application/json
   "new-reg": "https://example.com/acme/new-reg",
   "new-app": "https://example.com/acme/new-app",
   "revoke-cert": "https://example.com/acme/revoke-cert",
-  "revoke-cert-by-key": "https://example.com/acme/revoke-cert-by-key",
   "key-change": "https://example.com/acme/key-change",
   "meta": {
     "terms-of-service": "https://example.com/acme/terms",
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -306,36 +306,25 @@ JWS objects sent in ACME requests MUST meet the following additional criteria:
 * The JWS MUST NOT have the value "none" in its "alg" field
 * The JWS Protected Header MUST include the following fields:
   * "alg"
-  * "jwk"
+  * "kid" (except for new-reg and revoke-cert requests).
```
bifurcation: Just deleting "jwk" leaves a hole.  Don't you want something like the following?

```
  * "jwk" (only for requests to new-reg and revoke-cert resources)
  * "kid" (for requests to all resources except new-reg and revoke-cert)
```

These requirements should be reflected with MUSTs in the next paragraph.

draft-ietf-acme-acme.md
```diff
@@ -306,36 +306,25 @@ JWS objects sent in ACME requests MUST meet the following additional criteria:
...
   * "nonce" (defined below)
   * "url" (defined below)
 
-Note that this implies that GET requests are not authenticated.  Servers MUST
+The "kid" field should contain a registration URL received by POSTing to the
+new-reg resource. The "kid" field is not required when POSTing to the new-reg
+resource, because no such URL is available to the client yet. The "kid" field
+may be omitted for revoke-cert requests if they are signed by the
+certificate key rather than an account key.
```
bifurcation: This para should lay out the rules explicitly:
* For new-reg requests, there MUST be a `"jwk"` field containing the key being registered.
* For revoke-cert requests, there MUST be a`"jwk"` field containing a key authorized to revoke the cert.
* For all other requests, there MUST be a `"kid"` field containing the account URI

draft-ietf-acme-acme.md
```diff
@@ -1556,9 +1545,10 @@ Before revoking a certificate, the server MUST verify that the key used to sign
 the request is authorized to revoke the certificate.  The server SHOULD consider
 at least the following keys authorized for a given certificate:
 
-* the public key in the certificate.
+* The public key in the certificate. When using the public key in the
+  certificate, the client MUST omit the "kid" header field.
 
-* an account key that is authorized to act for all of the identifier(s) in the
+* An account key that is authorized to act for all of the identifier(s) in the
```
bifurcation: These are drive-bys, but I'll let them slide :)

draft-ietf-acme-acme.md
```diff
@@ -306,36 +306,25 @@ JWS objects sent in ACME requests MUST meet the following additional criteria:
...
+Note that authentication via signed POST implies that GET requests are not
+authenticated.  Servers MUST
 NOT respond to GET requests for resources that might be considered sensitive.
 
 In the examples below, JWS objects are shown in the JSON or flattened JSON
 serialization, with the protected header and payload expressed as
 base64url(content) instead of the actual base64-encoded value, so that the content
 is readable.  Some fields are omitted for brevity, marked with "...".
 
-## Equivalence of JWKs
```
bifurcation: Are there other places where we need comparison of JWKs?  As roll-over is currently specified, the CA needs to compare the "jwk" value in the header of the inner JWS to the "newKey" value in the payload.  If we're going to delete this section, we should consider fixing that, e.g., by removing the "jwk" from that JWS.

draft-ietf-acme-acme.md
jsha: Fixed.

draft-ietf-acme-acme.md
jsha: Fixed.

draft-ietf-acme-acme.md
jsha: Good point. I've left this section in for now, we can take another look at key roll-over separately.

jsha:
Updated based on feedback.

draft-ietf-acme-acme.md
```diff
@@ -303,11 +303,23 @@ JWS objects sent in ACME requests MUST meet the following additional criteria:
...
   * "nonce" (defined below)
   * "url" (defined below)
 
-Note that this implies that GET requests are not authenticated.  Servers MUST
+For new-reg requests, there MUST be a "jwk" field containing the key being
+registered, and there MUST NOT be a "kid" field.
+
+For revoke-cert requests that are authenticated by a certificate key, there
+MUST be a "jwk" field containing the public key from the relevant certificate,
+and there MUST NOT be a "kid" field.
```
bifurcation: Hm, "authenticated by a certificate key" makes me think:

* This para might be better stated as "For revoke requests, depends on the key being used to authorize.  Issuing / authorized account key => kid, cert key => jwk"
* We should require that if there is a "kid" field present, the server MUST ignore any "jwk" field that is present

draft-ietf-acme-acme.md
```diff
@@ -303,11 +303,23 @@ JWS objects sent in ACME requests MUST meet the following additional criteria:
...
+
+For all other requests, there MUST be a "kid" field containing the account URI
+received by POSTing to the new-reg resource.
+
+Note that authentication via signed POST implies that GET requests are not
+authenticated.  Servers MUST
```
bifurcation: Please re-wrap.

draft-ietf-acme-acme.md
```diff
@@ -922,11 +934,11 @@ Content-Type: application/jose+json
 {
   "protected": base64url({
     "alg": "ES256",
-    "jwk": {...},
```
bifurcation: It says above that you need a "jwk" for this request :)  It looks like you intend for the client to put the key in the "key" field, but I would be a little more comfortable having the CA assemble the object.  Basically for the same reason we are using "kid" in general -- because it shows forces the CA to process the thing.

draft-ietf-acme-acme.md
```diff
@@ -942,9 +954,9 @@ other fields that it does not recognize.  If new fields are specified in the
 future, the specification of those fields MUST describe whether they may be
 provided by the client.
 
-The server creates a registration object with the included contact information.
-The "key" element of the registration is set to the public key used to verify
-the JWS (i.e., the "jwk" element of the JWS header).  The server returns this
+The server creates a registration object with the included contact information
+and key.
+The server returns this
```
bifurcation: Same comment here.

jsha:
Updated. Rather than have a switch/case about how to authenticate revoke-cert, I split it into two resources, authenticated different ways: one by account, one by cert key.

draft-ietf-acme-acme.md
```diff
@@ -1544,16 +1556,10 @@ Content-Type: application/jose+json
...
-signed either with an account key pair or the key pair in the certificate.
 Before revoking a certificate, the server MUST verify that the key used to sign
 the request is authorized to revoke the certificate.  The server SHOULD consider
-at least the following keys authorized for a given certificate:
-
-* the public key in the certificate.
-
-* an account key that is authorized to act for all of the identifier(s) in the
-  certificate.
+authorized any account key that is authorized to act for all of the identifier(s) in the
```
cpu: This reads a little bit awkwardly to me. What about:

"The server SHOULD consider any account key authorized for all of the identifier(s) in the certificate to be authorized to revoke the certificate." ?

draft-ietf-acme-acme.md
```diff
@@ -303,12 +303,21 @@ JWS objects sent in ACME requests MUST meet the following additional criteria:
...
+  * "jwk" (only for requests to new-reg and revoke-cert resources)
+  * "kid" (for all requests except for new-reg and revoke-cert requests).
   * "nonce" (defined below)
   * "url" (defined below)
 
-Note that this implies that GET requests are not authenticated.  Servers MUST
-NOT respond to GET requests for resources that might be considered sensitive.
+For new-reg requests and revoke-cert-by-key requests, there MUST be a
+"jwk" field containing the key being registered, and there MUST NOT be a
+"kid" field.
```
bifurcation: Adding "revoke-cert-by-key" is too much to glom on to this PR.  Please pull it out and file a separate PR.  I think we can leave the remaining wording as it is (i.e., "jwk" for new-reg, "kid" for *), in anticipation of that follow-on.  No objection to "revoke-cert-by-key" in principle, I just think we might need an RTT or two on how exactly it's specified.

draft-ietf-acme-acme.md
```diff
@@ -303,12 +303,21 @@ JWS objects sent in ACME requests MUST meet the following additional criteria:
...
+
+For all other requests, there MUST be a "kid" field containing the account URI
+received by POSTing to the new-reg resource.
```
bifurcation: You still need to say what the server does when it receives a request with any combination of `kid` and `jwk`.  There are two possibilities:

1. Require the server to reject requests that don't meet `(kid && !jwk) || (!kid && jwk)`
2. Require the server to ignore `jwk` when `kid` is present.

I don't much care which.

draft-ietf-acme-acme.md
bifurcation: Also, please re-wrap.

draft-ietf-acme-acme.md
```diff
@@ -1577,6 +1583,71 @@ Content-Language: en
 }
 ~~~~~~~~~~
 
+## Certificate Revocation by Key
```
bifurcation: I don't think we need a whole new section for this, or at least we don't need quite so much copy/paste.  It seems like what's needed is just a paragraph or two that says "if you want to revoke using the key in the cert, do the same thing except (1) use a different endpoint, (2) sign with the key in the cert, and (3) use `jwk` instead of `kid`."

PR #194, opened 2016-09-26
Default to PEM with chain for certificates.

jsha:
One of the most common ACME deployment failures observed in practice is for
servers to be configured to serve only the end-entity certificate, without the
intermediate certificates. This is a particularly pernicious problem because
some browsers will still trust the resulting one-certificate chain, due to
caching or fetching of URLs from Authority Information Access. But other
browsers will not, resulting in a "works on my computer" problem.

Arguably this configuration is the result of incorrect clients, but we should
expect that most clients will do the easiest thing. This change aligns the
easiest thing with the most correct thing.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1285,7 +1285,8 @@ certificate URL.
 
 The default format of the certificate is PEM (application/x-pem-file),
 containing the end-entity certificate first, followed by any intermediate
-certificates that are needed to build a path to a trusted root. The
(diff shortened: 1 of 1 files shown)


martinthomson:
You should include advice here on omitting trust anchors.  TLS 1.3 has [some text](https://tlswg.github.io/tls13-spec/#rfc.section.4.4.1) you could crib from (see the definition of `certificate_list`).

seantek:
Hi, the PEM text is not correct: application/x-pem-file is not a registered media type, among other things. Will post to mailing list with details.

PR #195, opened 2016-09-27
Combine "requirements" and "authorizations."

jsha:
The previous definition of requirements created an awkward extra level of indirection, with status defined both on requirements and authorizations. It also created another object type, increasing the complexity of the protocol.

It was added mainly to provide a concept of an "out-of-band" requirement that could be satisfied, for instance, by adding funds to an account. However, this concept fits very well within the existing authorization framework, if we tweak the definition of authorizations so that authorizations are not always for domain names. Instead, we specify a type field on the authorization object, which can be "DNSName" or any other server-defined value. We can use the existing Out-of-Band challenge type to convey a link for users to visit to, for instance, provide payment.

For "DNSName" authorizations, the same constraints still hold - there is a defined DNS name, and the challenges apply to that specific name.

Editorially, I wound up moving the section about verifying correct Punycode encoding to new-application, since that is where the check needs to happen.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -721,11 +721,11 @@ notAfter (optional, string):
 : The requested value of the notAfter field in the certificate, in the date
 format defined in {{!RFC3339}}
 
-requirements (required, array):
+requirements (required, array of strings):
 : The requirements that the client needs to fulfill before the requested
 certificate can be granted (for pending applications).  For final applications,
-the requirements that were met.  Each entry is a dictionary with parameters
-describing the requirement (see below).
+the requirements that were met.  Each entry is the URL of an authorization
+object.
 
 certificate (optional, string):
 : A URL for the certificate that has been issued in response to this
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1207,6 +1153,9 @@ and MUST NOT issue a certificate with contents other than those requested.  If
 the server requires the request to be modified in a certain way, it should
 indicate the required changes using an appropriate error code and description.
 
+Servers MUST verify any identifier values that begin with the ASCII Compatible
+Encoding prefix "xn--" as defined in {{!RFC5890}} are properly encoded.
```
cpu: I think there's a slight error in the wording of this sentence. Should it read "as defined in {{!RFC5890}} **and** are properly encoded."?

jsha: Here's my parsing of it:

Servers Must verify any identifier values (that begin with the ASCII Compatible Encoding prefix "xn-- as defined in RFC5890) are properly encoded.

cpu: Ahhh! Ok, I see what you mean. Disregard my poor reading comprehension :-)

PR #200, opened 2016-10-27
Clarify that HTTP and TLS go over ports 80 and 443

### LINKED ISSUE #190: Explicitly limit to TCP ports 80 and 443
Since we did not get consensus around #4, we should be explicit that HTTP is TCP port 80, and HTTPS is TCP port 443, and that holds for redirects as well.

The is pretty much a minor wording change only.

bifurcation:
Fixes #190


cpu:
I spotted a typo/error introduced by this PR and submitted a follow-up: https://github.com/ietf-wg-acme/acme/pull/202

PR #202, opened 2016-10-28
Clarifies TLS port requirement for SNI challenge.

cpu:
Pull request #200 was recently merged to clarify that HTTP and TLS challenge requests must use port 80 and 443 respectively.

The "TLS with Server Name Indication (TLS SNI)" section was updated to read:

> Open a TLS connection to TCP port 443 the domain name being validated
> on the requested port, presenting SAN A in the SNI field.

This is a typo ("port 443 the domain name") and also still includes the phrase "on the requested port", which is no longer true.

This PR updates the SNI challenge description to fix the typo and to explicitly say that the server MUST use port 443 in the same style/manner as the HTTP edit from #200.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1838,13 +1838,13 @@ of the domain by verifying that the TLS server was configured appropriately,
 using these steps:
 
 1. Compute SAN A and SAN B in the same way as the client.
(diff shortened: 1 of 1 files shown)


bifurcation:
Thanks for fixing my lame typos 😄

PR #207, opened 2016-11-19
Change agreementRequired to userActionRequired

bifurcation:


draft-ietf-acme-acme.md
```diff
@@ -1086,7 +1086,7 @@ Content-Type: application/problem+json
 Content-Language: en
 
 {
-  "type": "urn:ietf:params:acme:error:agreementRequired"
+  "type": "urn:ietf:params:acme:error:userActionRequired"
   "detail": "Terms of service have changed"
```
cpu: Is there now a case beyond the terms of service changing that would require user action? Perhaps the `detail` message should be generalized slightly?

draft-ietf-acme-acme.md
```diff
@@ -482,7 +482,7 @@ to errors, this document defines the following standard tokens for use in the
 | invalidContact        | The contact URI for a registration was invalid     |
 | rejectedIdentifier    | The server will not issue for the identifier       |
 | unsupportedIdentifier | Identifier is not supported, but may be in future  |
-| agreementRequired     | The client must agree to terms before proceeding   |
+| userActionRequired    | The user must visit a URL before proceeding        |
```
cpu: I think changing "The user must visit a URL before proceeding" to "The user must perform an action specified by a URL before proceeding" would clarify this. As-is one might get the impression that a GET to the URL is sufficient and I think we're imagining something more involved.

draft-ietf-acme-acme.md
bifurcation: You can imagine other cases where you might want to break automation.  For example, if a user account has been dormant for a long time, and you want to do some advanced verification.  I think we can leave it up to CAs to determine in what cases they want to use this error.

However, according to [RFC 7807](https://tools.ietf.org/html/rfc7807#section-3.1), the `detail` field contains `human-readable explanation specific to this occurrence of the problem`.  So even if there are other use cases, it makes sense for this field to be specific to this use case.

PR #208, opened 2016-11-19
Remove the 'requirements' abstraction

bifurcation:
As discussed at IETF 97, this PR removes the "requirements" abstraction, and changes the "requirements" field in an application/order object to be an "authorizations" field.

This is a replacement for #195.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -189,7 +189,7 @@ Once the client is registered, there are three major steps it needs to take to
 get a certificate:
 
 1. Apply for a certificate to be issued
-2. Fulfill the server's requirements for issuance
+2. Prove control of any identifiers requested in the certificate
 3. Await issuance and download the issued certificate
 
 The client's application for a certificate describes the desired certificate
@@ -214,8 +214,8 @@ and make it available to the client.
 ~~~~~~~~~~
       Application
       Signature                     ------->
-                                    <-------            Requirements
-                                                  (e.g., Challenges)
+                                                             Required
+                                    <-------           Authorizations
 
       Responses
       Signature                     ------->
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -189,7 +189,7 @@ Once the client is registered, there are three major steps it needs to take to
 get a certificate:
 
 1. Apply for a certificate to be issued
-2. Fulfill the server's requirements for issuance
+2. Prove control of any identifiers requested in the certificate
```
cpu: "any identifiers" -> "all identifiers" ?

draft-ietf-acme-acme.md
```diff
@@ -760,11 +760,19 @@ notAfter (optional, string):
...
 format defined in {{!RFC3339}}
 
-requirements (required, array):
-: The requirements that the client needs to fulfill before the requested
-certificate can be granted (for pending applications).  For final applications,
-the requirements that were met.  Each entry is a dictionary with parameters
-describing the requirement (see below).
+authorizations (required, array):
+: For pending applications, the authorizations that the client needs to complete
+before the requested certificate can be granted (see
```
cpu: I think changing "granted" -> "issued" would better match the wording used elsewhere.

draft-ietf-acme-acme.md
```diff
@@ -760,11 +760,19 @@ notAfter (optional, string):
...
-describing the requirement (see below).
+authorizations (required, array):
+: For pending applications, the authorizations that the client needs to complete
+before the requested certificate can be issued (see
+{{identifier-authorization}}).  For final applications, the authorizations that
+were completed.  Each entry is a dictionary with parameters describing the
+authorization:
+
+  status (required, string):
+  : The status of the authorization.  This field MUST have the same value as it
```
jsha: I thought we were going to make this a simple array of URLs rather than inlining the status (and therefore having it in two places)?

PR #209, opened 2016-11-19
Add an external account binding field

### LINKED ISSUE #170: add a special token parameter in registration

bifurcation:
As discussed at IETF 97, we've agreed to have an external account binding field, but we need to have some more security considerations.  r? @martinthomson for the security considerations.

Replacement for #172 

Fixes #170

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -677,6 +677,10 @@ contact (optional, array of string):
 related to this authorization. For example, the server may wish to notify the
 client about server-initiated revocation.
 
+external-account-token (optional, string):
+: A secret value provided to the client out of band that associates this account
+with another account in a different system.
+
 terms-of-service-agreed (optional, boolean):
 : Including this field in a new-registration request, with a value of true,
 indicates the client's agreement with the terms of service. This field is not
@@ -698,6 +702,7 @@ updateable by the client.
     "mailto:cert-admin@example.com",
     "tel:+12025551212"
   ],
+  "external-account-token": "rAPsLE6hxgut8O6fu2Bj3A",
   "terms-of-service-agreed": true,
   "applications": "https://example.com/acme/reg/1/apps"
 }
@@ -952,7 +957,8 @@ resource, in order to prevent caching of this resource.
 
 A client creates a new account with the server by sending a POST request to the
 server's new-registration URI.  The body of the request is a stub registration
-object containing only the "contact" field.
+object containing only fields that can be set by the client (e.g., "contact",
+"external-account-token", "terms-of-service-agreed").
 
 ~~~~~~~~~~
 POST /acme/new-reg HTTP/1.1
@@ -968,6 +974,7 @@ Content-Type: application/jose+json
   })
   "payload": base64url({
     "terms-of-service-agreed": true,
+    "external-account-token": "rAPsLE6hxgut8O6fu2Bj3A",
     "contact": [
       "mailto:cert-admin@example.com",
       "tel:+12025551212"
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1000,6 +1007,20 @@ then it MUST return a 200 (OK) response and provide the URI of that registration
 in a Content-Location header field.  This allows a client that has an account
 key but not the corresponding registration URI to recover the registration URI.
 
+The server MAY require a value to be present for the "external-account-token"
+field.  This can be used to an ACME account with an existing account in a
```
martinthomson: missing word here "used to [associate|bind] an ACME account"

draft-ietf-acme-acme.md
```diff
@@ -2597,6 +2618,33 @@ within their trusted network and use these resolvers both for both CAA record
 lookups and all record lookups in furtherance of a challenge scheme (A, AAAA,
 TXT, etc.).
 
+## External Accounts
```
martinthomson: The discussion in Seoul also mentioned the use of ACME account keys to bind existing accounts together.  That is, where the CA has forms that allow a user to enter details of their ACME accounts. 

I think that we want advice on how to construct a token that proves to a CA that the account key is controlled by the current user.  This is a new use of the account key outside the ACME protocol, which is potentially risky if the user can be convinced to sign an arbitrary challenge.  (Cross domain usage, etc...)  Advice on construction is probably useful.  Standardizing the form would be better.

draft-ietf-acme-acme.md
```diff
@@ -2597,6 +2618,33 @@ within their trusted network and use these resolvers both for both CAA record
...
+the CA structures its tokens.
+
+If the CA simply provides the client with a random value to use in this field,
+then there is a risk that the value will be stolen and misused.  For example, if
+the CA uses a CDN to terminate HTTPS for ACME, then the CDN could steal a token
+and use it to bind real customer's account to the CDN's ACME account (instead of
+the real customer's ACME account).  CAs that construct tokens in this way SHOULD
+avoid the use of intermediaries that terminate HTTPS.
+
+CAs can mitigate this risk by reducing the privilege associated with the token.
```
martinthomson: "this risk" is pretty vague by now.  Please expand.

draft-ietf-acme-acme.md
```diff
@@ -2597,6 +2618,33 @@ within their trusted network and use these resolvers both for both CAA record
...
+Instead of granting the holder of the token the full privileges associated with
+the account, for example, the CA could simply use the association for
+record-keeping, so that the external account holder would be notified of actions
+by the ACME client, and the ACME client would still have to acquire privileges
+on its own (proving control of identifiers, etc.).
```
martinthomson: I think that I would remove this paragraph given the specific comment below.

draft-ietf-acme-acme.md
```diff
@@ -2597,6 +2618,33 @@ within their trusted network and use these resolvers both for both CAA record
...
+
+CAs SHOULD construct tokens in a way that binds them to a particular ACME
+account key pair.  For example, a CA might use JWTs containing a thumbprint of
+the account public key {{?RFC7519}}.  These tokens can safely be used even in
+the presence of HTTPS-terminating intermediaries, since they can only be used by
+the holder of the corresponding account private key.
```
martinthomson: Note that this design is only safe because the request bearing the external account token is covered by a signature.

prosecco:
As I just wrote on the ACME mailing list, let's not use a bearer "external-token"

For example, a much safer way to embed “external_secret”, “ca-account-secret”, and “token” would be to have an “external-key-id” field, that indicates a MAC key that has been exchanged out-of-band; we would then use MAC-based authentication to bind the MAC key to the relevant JWS request, using the standard nested authentication mechanism we are using in roll-over, something like:
	
	Sign (account-key, (request, external-key-id, MAC (external_key, Hash(account_key))))

This would ensure that the external secret is never sent on the wire and is not a bearer token,
and the compound authentication guarantees of the resulting protocol are quite strong.

bifurcation:
Closing in favor of #212

PR #212, opened 2016-11-22
Add MAC-based external account binding

bifurcation:
This is a variant of #209 that uses a MAC key to perform the account binding instead of a bearer token.  Thanks to @prosecco for pointing out the need for this on the ACME mailing list.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -682,6 +682,11 @@ terms-of-service-agreed (optional, boolean):
 indicates the client's agreement with the terms of service. This field is not
 updateable by the client.
 
+external-account-binding (optional, JWS):
+: Including this field in a new-registration request indicates that the client
+wishes to associate this account with an account in some external system, using
+the process described in {{external-account-binding}}
+
 applications (required, string):
 : A URI from which a list of authorizations submitted by this account can be
 fetched via a GET request.  The result of the GET request MUST be a JSON object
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1092,6 +1097,75 @@ Content-Language: en
...
+  }),
+  "signature": "5TWiqIYQfIDfALQv...x9C2mg8JGPxl5bI4"
+}
+~~~~~
+
+When a CA receives a new-registration request containing an
+"external-account-binding" field, it MUST perform the following steps:
+
+1. Verify that the value of the field is a well-formed JWS
+2. Verify that the JWS protected meets the above criteria
```
clintwilson: 2. Verify that the JWS protected header meets the above criteria

draft-ietf-acme-acme.md
```diff
@@ -1092,6 +1097,75 @@ Content-Language: en
 }
 ~~~~~
 
+### External Account Binding
+
+The server MAY require a value to be present for the "external-account-binding"
+field.  This can be used to an ACME account with an existing account in a
```
clintwilson: field.  This can be used to bind an ACME account with an existing account in a

martinthomson:
Question: will you require a CA that doesn't use external account bindings to reject a request if it contains an external-account-binding member?

nit: s/field/member/g, consistent with RFC 7159.

draft-ietf-acme-acme.md
```diff
@@ -1092,6 +1097,75 @@ Content-Language: en
...
+      }),
+      "payload": base64url(/* same as in "jwk" above */),
+      "signature": /* MAC using MAC key from CA */
+    }
+  }),
+  "signature": "5TWiqIYQfIDfALQv...x9C2mg8JGPxl5bI4"
+}
+~~~~~
+
+When a CA receives a new-registration request containing an
```
jsha: This should contain a provision allowing CAs to ignore the external-account-binding field if they do not support it. E.g.

"When a CA receives a new-registration request containing an external-account-binding field, if the CA supports that field it MUST..."

PR #218, opened 2016-11-29
Add an error code for unsupported signature algorithm

### LINKED ISSUE #217: Enable server to signal unsupported signature algorithm
The server needs to support whatever the client sends.  We should have an error code for this, and a way for the server to specify what algorithms it does support in the error message.

bifurcation:
Also:

* Forbid the use of MAC-based JWS algorithms (in case this wasn't obvious)
* Sort error codes alphabetically

Fixes #217


bifurcation:
r+ from @martinthomson via XMPP

PR #221, opened 2016-12-02
Fix registry entry for "external-account-token"

bifurcation:
Failure to synchronize between #211 and #212.

Also removes a stray word and some whitespace in the vicinity.


uhhhh2:
Also in the registries, is the "authorizations" property of an order object an array of strings (URLs) or an array of objects (each of which contains a "URL" parameter)?

bifurcation:
@uhhhh2 array of strings, since #208

PR #223, opened 2016-12-04
Cleanup authorizations field of order object  

uhhhh2:
Removed unnecessary paragraphs from authorizations field description in order object.


bifurcation:
Thanks @uhhhh2, not sure how I missed that in #208 :)

PR #230, opened 2017-01-17
key-change: update example according to inner jws rules

Trim:
Rules above cleary say that inner JWS has this requirement:
> * The inner JWS MUST have the same "url" parameter as the outer JWS.

Server request validation clearly needs this parameter too in the inner JWS:
> 5. Check that the "url" parameters of the inner and outer JWSs are the same


bifurcation:
Merging based on manual review; CircleCI seems to be busted at the moment.

PR #231, opened 2017-01-19
Clarify initial section.

jsha:
Several places talked about account key pairs as being the primary notion of an
account, while the account is now primary and has a public key associated with
it. Updated to clarify that.

Clarified that the account request can contain more than just contact
information.

Fixed the order of operations in protocol summary to reflect new-order flow.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -224,9 +224,8 @@ and make it available to the client.
                                     <-------             Certificate
 ~~~~~~~~~~
 
-To revoke a certificate, the client sends a revocation request indicating
-the certificate to be revoked, signed with the account's private key, or with
-the private key corresponding to the public key in the certificate.
+To revoke a certificate, the client sends a signed revocation request indicating
+the certificate to be revoked:
 
 ~~~~~~~~~~
       Client                                                 Server
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -160,33 +160,31 @@ from the client.
 # Protocol Overview
 
 ACME allows a client to request certificate management actions using a set of
-JSON messages carried over HTTPS.   In some ways, ACME functions much like a
-traditional CA, in which a user creates an account, adds identifiers to that
-account (proving control of the domains), and requests certificate issuance for
-those domains while logged in to the account.
+JSON messages carried over HTTPS.   In many ways, ACME functions much like a
+traditional CA, in which a user creates an account, requests a certificate,
+and proves control of the domains in that certificate in order for the CA to
+sign the requested certificate.
 
-In ACME, the account is represented by an account key pair.  The "add a domain"
-function is accomplished by authorizing the key pair for a given domain.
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -226,9 +224,9 @@ and make it available to the client.
                                     <-------             Certificate
 ~~~~~~~~~~
 
-To revoke a certificate, the client simply sends a revocation request indicating
-the certificate to be revoked, signed with an authorized key pair. The server
-indicates whether the request has succeeded.
+To revoke a certificate, the client sends a revocation request indicating
+the certificate to be revoked, signed with the account's private key, or with
```
cpu: I like your new wording more, but I think there's one aspect lost in translation.

The text "signed with an authorized key pair" allows for the notion that I can create a 2nd account keypair, authorize all the same domains and have that new account keypair considered an "authorized key pair" for revoking a certificate with the same set of names. 

This is a pretty clunky thing to express but I think the new version makes it sound like only the original issuing account keypair or the certificate's private key could be used for revocation.

Am I misunderstanding?

draft-ietf-acme-acme.md
```diff
@@ -160,33 +160,31 @@ from the client.
...
-JSON messages carried over HTTPS.   In some ways, ACME functions much like a
-traditional CA, in which a user creates an account, adds identifiers to that
-account (proving control of the domains), and requests certificate issuance for
-those domains while logged in to the account.
+JSON messages carried over HTTPS.   In many ways, ACME functions much like a
+traditional CA, in which a user creates an account, requests a certificate,
+and proves control of the domains in that certificate in order for the CA to
+sign the requested certificate.
 
-In ACME, the account is represented by an account key pair.  The "add a domain"
```
bifurcation: I like this paragraph, because it clarifies how ACME maps to existing CA operations.  Why are you deleting it?

richsalz: Yes, that's not strictly editorial; it's important.

jsha: Because it's no longer true that the account is represented by a key pair. An account is represented by a URL; it has a key pair as a property. I'm happy to entertain suggestions for a rewrite of this paragraph, but when I tried, it wound up like "In ACME, there are accounts." Which seemed uninformative.

jsha: Or rather, "In ACME, there are accounts" is redundant with the next paragraph:

> The first phase of ACME is for the client to register with the ACME server. The client generates an asymmetric key pair and associates this key pair with a set of contact information by signing the contact information. The server acknowledges the registration by replying with a account object echoing the client's inputrequest an account with the ACME server. The client generates an asymmetric key pair and requests a new account, optionally providing contact information, agreeing to terms of service, and/or associating the account with an existing account in another system. The creation request is signed with the generated private key to prove that the client controls it.

I think that paragraph does an adequate job describing how ACME accounts are similar to traditional CA accounts. Do you disagree?

PR #232, opened 2017-01-19
Editorial fixes round 2

jsha:
Fix reference to old, partial, URL integrity mechanism.

Remove line about distinct URLs for singular resources, as there is no mechanism
for this now.

Standardize on `new-account` rather than `new-acct`; previously there was a mix.

Replace /acme/acct/asdf in examples with /acme/acct/1.

Fix a few miscellaneous errors.

Renumber a list to use correct numbers.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -520,7 +520,7 @@ enables:
 
   * Account Creation
   * Ordering a Certificate
-  * Identifiear Authorization
+  * Identifier Authorization
   * Certificate Issuance
   * Certificate Revocation
 
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -374,10 +374,8 @@ requests for denial-of-service protection.
 
 These intermediaries can also change values in the request that are not signed
 in the HTTPS request, e.g., the request URI and headers.  ACME uses JWS to
-provide a limited integrity mechanism, which protects against an intermediary
-changing the request URI to another ACME URI of a different type.  (It does not
-protect against changing between URIs of the same type, e.g., from one
-authorization URI to another).
+provide an integrity mechanism, which protects against an intermediary
+changing the request URI to another ACME URI.
 
 As noted above, all ACME request objects carry a "url" parameter in their
 protected header.  This header parameter encodes the URL to which the client is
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -519,9 +518,9 @@ indicate why the server was unable to validate authorization.
 In this section, we describe the certificate management functions that ACME
 enables:
 
-  * Account Key Registration
+  * Account Creation
   * Ordering a Certificate
-  * Account Key Authorization
+  * Identifiear Authorization
```
cpu: Typo: "Identifiear" -> "Identifier"

jsha:
My plan is to replace other uses of `asdf` in subsequent PRs.

kelunik:
> Standardize on new-account rather than new-acct; previously there was a mix.

Now `new-authz` is the only abbreviated one as far as I can see, so that should probably adjusted to `new-authorization`, too?

draft-ietf-acme-acme.md
```diff
@@ -597,15 +592,15 @@ establish a new account with the server, prove control of an identifier, issue a
...
-| Poll for status      | GET  order     | 200          |
-| Check for new cert   | GET  cert      | 200          |
+| Action               | Request           | Response        |
+|:---------------------|:------------------|:----------------|
+| Get a nonce          | HEAD new-nonce    | 204             |
+| Create account       | POST new-account  | 201 -> account  |
+| Submit an order      | POST new-order    | 201 -> order    |
+| Fetch challenges     | GET  authz        | 200             |
+| Respond to challenge | POST challenge    | 200             |
+| Poll for status      | GET  authz        | 200             |
```
bifurcation: It seems like you should poll the order object here, because that's how you get the certificate URL.  Otherwise you have to go fetch the order once the authz turns good.

draft-ietf-acme-acme.md
```diff
@@ -1119,7 +1116,7 @@ Content-Type: application/jose+json
 }
 ~~~~~
 
-When a CA receives a new-registration request containing an
+When a CA receives a new-account request containing an
```
bifurcation: Nit: Re-wrap might be appreciated

draft-ietf-acme-acme.md
```diff
@@ -1259,7 +1256,7 @@ provide a way to reactivate a deactivated account.
 
 ## Applying for Certificate Issuance
 
-The holder of an account key pair may use ACME to submit an order for a
+A client may use ACME to submit an order for a
```
bifurcation: Nit: Re-wrap

draft-ietf-acme-acme.md
jsha: I avoid re-wrapping unless a paragraph is very badly out of alignment, because it destroys `git blame`. It also makes diffs harder to read. Any objection?

draft-ietf-acme-acme.md
jsha: That's incidental to this change, which just re-tabulates the table. Filed a separate issue: https://github.com/ietf-wg-acme/acme/issues/237

PR #234, opened 2017-01-19
Remove SCT link relation.

jsha:
This is unlikely to be implemented by any CA, and there is a good
substitute available: servers that want to provide SCTs in a TLS
extension can submit their own certificates. However, even that
will be exceedingly rare as most CAs are planning to embed SCTs
in certificates.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,16 +21,16 @@ meaning of the specification), you can either:
   a) Fork this repository and submit a pull request; this is the lowest
   friction way to get editorial changes in.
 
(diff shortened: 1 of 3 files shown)


bifurcation:
This needs WG review before merging.  I don't disagree in principle, especially because it's easy to add back later.  But if there is someone out there who plans to use this, it's not much extra spec.

jsha:
WG review == at IETF meeting, or on mailing list?

richsalz:
Mailing list is where all work is 'officially' done.

bifurcation:
This appears to already have been done.  There is no "sct" link relation in current master.

PR #235, opened 2017-01-19
More editorial fixes.

jsha:
Include "processing" as a possible status for an order (this was listed
elsewhere but not consistent).

Include a tls-sni-02 challenge in example authz.

Clarify that finalizing an authorization involves *modifying* the expires field,
not adding it.

Clarify in revocation section the two types of accounts that can request a
certificate revocation.

Expand examples in challenge section to match format in the rest of the doc,
including HTTP headers and JWS wrapping.

Indicate for HTTP, TLS SNI, and DNS, that making multiple requests is not only
supported, but RECOMMENDED.

Fix a mismatch between "client configurable" in description of registries and
"configurable" in the table of things to be registered.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1583,7 +1583,7 @@ Link: <https://example.com/acme/some-directory>;rel="directory"
       "type": "tls-sni-02",
       "url": "https://example.com/authz/1234/1",
       "token": "DGyRejmCefe7v4NfDGDKfA"
-    }
+    },
     {
       "type": "dns-01",
       "url": "https://example.com/authz/1234/2",
@@ -2173,7 +2173,7 @@ Content-Type: application/jose+json
     "alg": "ES256",
     "kid": "https://example.com/acme/acct/1",
     "nonce": "JHb54aT_KTXBWQOzGYkt9A",
-    "url": "https://example.com/acme/authz/1234/1"
+    "url": "https://example.com/acme/authz/1234/2"
   }),
   "payload": base64url({
     "keyAuthorization": "evaGxfADs...62jcerQ"
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1645,7 +1645,8 @@ that have been provisioned to a web server.
 The server is said to "finalize" the authorization when it has completed
 one of the validations, by assigning the authorization a status of "valid"
 or "invalid", corresponding to whether it considers the account authorized
-for the identifier. When finalizing an authorization, the server MAY remove
+for the identifier. If the final state is "valid", then the server MUST include
+an "expires" field. When finalizing an authorization, the server MAY remove
 challenges other than the one that was completed, and may modify the "expires"
 field. The server SHOULD NOT remove challenges with status "invalid".
 
@@ -1767,12 +1768,13 @@ Before revoking a certificate, the server MUST verify that the key used to sign
 the request is authorized to revoke the certificate.  The server SHOULD consider
 at least the following accounts authorized for a given certificate:
 
-* the public key in the certificate.
-
 * the account that issued the certificate.
 
 * an account that holds authorizations for all of the identifiers in the
   certificate.
+
+The server SHOULD also consider a revocation request valid if it is signed with
+the private key corresponding to the public key in the certificate.
 
 If the revocation succeeds, the server responds with status code 200 (OK).  If
 the revocation fails, the server returns an error.
@@ -1996,11 +1998,6 @@ domain by verifying that the resource was provisioned as expected.
    server SHOULD ignore whitespace characters at the end of the body.
 5. Verify that key authorization provided by the server matches the token for
    this challenge and the client's account key.
-
-The server MAY follow redirects when dereferencing the URI.
-
-It is RECOMMENDED that the server dereference the URI multiple times from
-various network perspectives, in order to make MitM attacks harder.
 
 If all of the above verifications succeed, then the validation is successful.
 If the request fails, or the body does not pass these checks, then it has
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -751,8 +751,8 @@ the server requires the client to complete, and any certificates that have
 resulted from this order.
 
 status (required, string):
-: The status of this order.  Possible values are: "pending", "valid", and
-"invalid".
+: The status of this order.  Possible values are: "pending", "processing",
+"valid", and "invalid".
 
 expires (optional, string):
 : The timestamp after which the server will consider this order invalid, encoded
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1575,12 +1576,17 @@ Link: <https://example.com/acme/some-directory>;rel="directory"
...
-      "url": "https://example.com/authz/asdf/0",
-      "token": "IlirfxKKXAsHtmzK29Pj8A"
+      "url": "https://example.com/authz/1234/0",
+      "token": "DGyRejmCefe7v4NfDGDKfA"
     },
     {
+      "type": "tls-sni-02",
+      "url": "https://example.com/authz/1234/1",
+      "token": "DGyRejmCefe7v4NfDGDKfA"
+    }
```
cpu: missing a `,` here.

draft-ietf-acme-acme.md
```diff
@@ -2111,11 +2164,22 @@ keyAuthorization (required, string):
...
+Host: example.com
+Content-Type: application/jose+json
+
 {
-  "keyAuthorization": "evaGxfADs...62jcerQ"
+  "protected": base64url({
+    "alg": "ES256",
+    "kid": "https://example.com/acme/acct/1",
+    "nonce": "JHb54aT_KTXBWQOzGYkt9A",
+    "url": "https://example.com/acme/authz/1234/1"
```
cpu: I think this `"url"` should be `https://example.com/acme/authz/1234/2` instead of `/1`. Yes?

draft-ietf-acme-acme.md
```diff
@@ -1639,10 +1645,9 @@ that have been provisioned to a web server.
 The server is said to "finalize" the authorization when it has completed
 one of the validations, by assigning the authorization a status of "valid"
 or "invalid", corresponding to whether it considers the account authorized
-for the identifier.  If the final state is "valid", the server MUST add an
-"expires" field to the authorization.  When finalizing an authorization,
-the server MAY remove challenges other than the one that was completed. The
-server SHOULD NOT remove challenges with status "invalid".
+for the identifier. When finalizing an authorization, the server MAY remove
+challenges other than the one that was completed, and may modify the "expires"
```
bifurcation: The intent here is to say that an authorization in the "valid" state MUST have an "expires" field.  So we should not remove this entirely.  How about:

```
If the final state is "valid", then the server MUST include an "expires" field in the authorization.
```

That doesn't presume that one was not already present.

draft-ietf-acme-acme.md
```diff
@@ -1759,11 +1765,13 @@ Revocation requests are different from other ACME request in that they can be
 signed either with an account key pair or the key pair in the certificate.
 Before revoking a certificate, the server MUST verify that the key used to sign
 the request is authorized to revoke the certificate.  The server SHOULD consider
-at least the following keys authorized for a given certificate:
+at least the following accounts authorized for a given certificate:
```
bifurcation: Problem with this is that the public key in the certificate might not have an account.  You could refactor this to say "The following accounts: [issued, authorized]  The server SHOULD also consider a revocation request valid if it is signed with the private key corresponding to the public key in the certificate."

draft-ietf-acme-acme.md
```diff
@@ -1916,8 +1924,14 @@ It MUST NOT contain any characters outside the URL-safe Base64 alphabet and MUST
 NOT contain any padding characters ("=").
 
 ~~~~~~~~~~
+GET /acme/authz/1234/0 HTTP/1.1
```
bifurcation: Including all the HTTP and JWS glue in here is counter-productive.  The point of these sections is to highlight the structure of the object you need to send, not the HTTP/JWS mechanics.  The HTTP/JWS stuff obscures that message.

draft-ietf-acme-acme.md
```diff
@@ -1972,6 +1997,11 @@ domain by verifying that the resource was provisioned as expected.
 5. Verify that key authorization provided by the server matches the token for
    this challenge and the client's account key.
 
+The server MAY follow redirects when dereferencing the URI.
```
bifurcation: This needs WG review.   Exempting servers from standard HTTP semantics is novel, and it has serious operational implications for servers, since there are lots of solutions for things like load-balanced servers that rely on redirects.

draft-ietf-acme-acme.md
```diff
@@ -1972,6 +1997,11 @@ domain by verifying that the resource was provisioned as expected.
...
+
+It is RECOMMENDED that the server dereference the URI multiple times from
+various network perspectives, in order to make MitM attacks harder.
```
bifurcation: Same comment here.  New normative language, and it implies that server's can't just take down the verification thing after they get a query, but rather keep it up for some indeterminate period of time.

draft-ietf-acme-acme.md
jsha: So your intent is that the server MUST follow redirects? I'm fine with that, just trying to clarify the spec since it doesn't say either way.

draft-ietf-acme-acme.md
jsha: The same is true of all the other example sections, where you added the obscuring HTTP/JWS glue. I happen to agree that the glue is distracting, but we should be consistent about including or not including it.

draft-ietf-acme-acme.md
jsha: Fair enough. Will break this out into a separate PR, and clarify that the time is not indeterminate, but merely until the authz transitions to valid or invalid.

draft-ietf-acme-acme.md
jsha: Good call, fixed.

draft-ietf-acme-acme.md
jsha: Fixed, thanks.

draft-ietf-acme-acme.md
jsha: Actually, MAY matches the [HTTP RFC](https://tools.ietf.org/html/rfc7231#section-6.4):

> If a Location header field (Section 7.1.2) is provided, the user agent MAY automatically redirect its request to the URI referenced by the Location field value

draft-ietf-acme-acme.md
```diff
@@ -1759,13 +1766,16 @@ Revocation requests are different from other ACME request in that they can be
...
+at least the following accounts authorized for a given certificate:
 
-* the public key in the certificate.
+* the account that issued the certificate.
 
-* an account key that is authorized to act for all of the identifier(s) in the
+* an account that holds authorizations for all of the identifiers in the
   certificate.
 
+The server SHOULD also consider a revocation request valid if it is signed with
```
kelunik: Any reason why this is a SHOULD and not a MUST?

jsha: There may be other CAs that don't want to consider possession of a private key sufficient for automated revocation. For instance they might want to have people submit a certificate problem report and manually validate it. I wanted to lean on the side of not tying their hands.

kelunik: @jsha But how is this related to the certificate key? If I have the account key, I can still always revoke it? If revocation with the certificate key is an optional feature, I see little sense to have it at all, because clients will then rather choose the "safe" (always working) variant and use the account key.

PR #238, opened 2017-01-23
Specify server MAY follow HTTP redirects.

jsha:
Per HTTP RFC: https://tools.ietf.org/html/rfc7231#section-6.4

See prior discussion at https://github.com/ietf-wg-acme/acme/pull/235#discussion_r97092394.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1972,6 +1972,8 @@ domain by verifying that the resource was provisioned as expected.
 5. Verify that key authorization provided by the server matches the token for
    this challenge and the client's account key.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1985,6 +1985,8 @@ domain by verifying that the resource was provisioned as expected.
 5. Verify that key authorization provided by the HTTP server matches the token
    for this challenge and the client's account key.
 
+The server MAY follow redirects when dereferencing the URI.
```
bifurcation: How about `SHOULD`?  It seems like having a clear expectation here is in the interest of interoperability.

PR #239, opened 2017-01-23
Specify multi-viewpoint validation.

jsha:
Servers may make multiple validation requests; clients should keep resources
around until authz changes.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1930,7 +1930,7 @@ for the domain in question.
 The path at which the resource is provisioned is comprised of the fixed prefix
 ".well-known/acme-challenge/", followed by the "token" value in the challenge.
 The value of the resource MUST be the ASCII representation of the key
-authorization. The resource should remain provisioned there until the client
+authorization. The resource should remain provisioned until the client
 observes that the authorization resource has a status of either "valid" or
 "invalid."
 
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1930,7 +1930,9 @@ for the domain in question.
 The path at which the resource is provisioned is comprised of the fixed prefix
 ".well-known/acme-challenge/", followed by the "token" value in the challenge.
 The value of the resource MUST be the ASCII representation of the key
-authorization.
+authorization. The resource should remain provisioned there until the client
```
cpu: tiny editorial nit: I think "there" in this sentence reads a little funny and is redundant with the initial path sentence. Can you say "The resource should remain provisioned until the client ...." and leave it at that? If you disagree I'd let it rest as-is :-)

draft-ietf-acme-acme.md
```diff
@@ -2630,11 +2635,6 @@ perform all necessary checks before issuing.
 
 CAs using ACME to allow clients to agree to terms of service should keep in mind
 that ACME clients can automate this agreement, possibly not involving a human
-user.  If a CA wishes to have stronger evidence of user consent, it may present
-an out-of-band requirement or challenge to require human involvement.
-
-# Operational Considerations
-
```
bifurcation: Accidental deletion?  No longer really makes sense with this chunk removed.

jsha:
Closing in favor of #260.

PR #240, opened 2017-01-30
Discuss invalidation of authz after compromise.

jsha:


draft-ietf-acme-acme.md
```diff
@@ -2688,6 +2691,23 @@ within their trusted network and use these resolvers both for both CAA record
...
+issued, and to ensure that the attacker cannot issue new certificates for the
+affected hostnames.
+
+If the attacker gained access to the private key for an ACME account, but has
+not yet changed the ACME account keypair, the proper owner can use the key
+rotation functionality to regain exclusive control of their ACME account.
+
+If the attacker did change the ACME account keypair, or if the attacker created
+their own ACME account and authorized the hostname, the proper owner, after
+regaining control of the host, can authorize that hostname on their own account
```
cpu: I think some potential ambiguity could be removed by saying "can authorize that hostname on their own *new* account" here.

bifurcation:
I understand the risk you're talking about here, but taking an implicit approach to solving it is wrong.   For example:

* The client has no way to know whether the server has obeyed the SHOULD here
* There are use cases for multiple accounts being authorized for the same identifier (think about things like multi-CDN)

If we're going to include a mechanism in this document for recovering from account compromise, then I would propose that it take the form of a "reset this domain" endpoint.  That is, enable the client to explicitly direct the server to revoke all authorizations for a domain except for some known set:

```
POST /acme/reset-identifier HTTP/1.1

JWS({
  "identifier": {
    "type": "dns",
    "value": "victim.com"
  },
  "except": [
    "https://example.com/acme/authz/for-account-1",
    "https://example.com/acme/authz/for-account-2"
  ]
})
```

Given that this is a pretty significant new feature, at this point, I'm inclined to hold it for a follow-on spec.  Just one more endpoint in the `directory`.

jsha:
Agreed that this is fairly complicated use case. Needs more discussion, and doesn't need to be part of this revision of the spec.

PR #251, opened 2017-02-14
Clarify that servers must reject disallowed revocation reasonCodes

rolandshoemaker:
Instead of leaving it ambiguous.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -493,24 +493,25 @@ information using problem document {{!RFC7807}}.  To facilitate automatic
 response to errors, this document defines the following standard tokens for use
 in the "type" field (within the "urn:ietf:params:acme:error:" namespace):
 
-| Code                  | Description                                                        |
-|:----------------------|:-------------------------------------------------------------------|
-| badCSR                | The CSR is unacceptable (e.g., due to a short key)                 |
-| badNonce              | The client sent an unacceptable anti-replay nonce                  |
-| badSignatureAlgorithm | The JWS was signed with an algorithm the server does not support   |
-| caa                   | CAA records forbid the CA from issuing                             |
-| connection            | The server could not connect to validation target                  |
-| dnssec                | DNSSEC validation failed                                           |
-| invalidContact        | The contact URI for a registration was invalid                     |
-| malformed             | The request message was malformed                                  |
-| rateLimited           | The request exceeds a rate limit                                   |
-| rejectedIdentifier    | The server will not issue for the identifier                       |
-| serverInternal        | The server experienced an internal error                           |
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1751,7 +1751,8 @@ reason (optional, int):
 to be used when generating OCSP responses and CRLs. If this field is not set
 the server SHOULD use the unspecified (0) reasonCode value when generating OCSP
 responses and CRLs. The server MAY disallow a subset of reasonCodes from being
-used by the user.
+used by the user. If a request contains a disallowed reasonCode the server MUST
+it.
```
jsha: missing "reject"

draft-ietf-acme-acme.md
```diff
@@ -1751,7 +1751,8 @@ reason (optional, int):
 to be used when generating OCSP responses and CRLs. If this field is not set
 the server SHOULD use the unspecified (0) reasonCode value when generating OCSP
 responses and CRLs. The server MAY disallow a subset of reasonCodes from being
-used by the user.
+used by the user. If a request contains a disallowed reasonCode the server MUST
+reject it.
```
jsoref: I'd request that there be a defined Error code for this case.

draft-ietf-acme-acme.md
```diff
@@ -1752,7 +1753,9 @@ to be used when generating OCSP responses and CRLs. If this field is not set
 the server SHOULD use the unspecified (0) reasonCode value when generating OCSP
 responses and CRLs. The server MAY disallow a subset of reasonCodes from being
 used by the user. If a request contains a disallowed reasonCode the server MUST
-reject it.
+reject it with the error type "urn:ietf:params:acme:error:badRevocationReason".
+The problem document returned with the error MUST include an "supportedReasons"
```
jsoref: an => a

draft-ietf-acme-acme.md
```diff
@@ -1751,7 +1752,10 @@ reason (optional, int):
 to be used when generating OCSP responses and CRLs. If this field is not set
 the server SHOULD use the unspecified (0) reasonCode value when generating OCSP
 responses and CRLs. The server MAY disallow a subset of reasonCodes from being
-used by the user.
+used by the user. If a request contains a disallowed reasonCode the server MUST
+reject it with the error type "urn:ietf:params:acme:error:badRevocationReason".
+The problem document returned with the error MUST include a "supportedReasons"
```
jsha: I'm reluctant to add extra infrastructure (the "supportedReasons" field) for a use case that will be exceedingly rare. We don't really expect most clients to implement some sort of fancy menu like "Try one of these other revocation reasons!" Instead how about we just say "the problem detail SHOULD indicate which revocation reasons are allowed."

PR #255, opened 2017-02-15
agreement: tautology 4=4

jsoref:


jsha:
Note is still required.

PR #256, opened 2017-02-15
spelling: rfc-editor

jsoref:
This is more of a proposal.

Grammarly doesn't like "RFC EDITOR", I think I agree, but...


jsha:
I have a feeling this may be an IETF style thing. I don't have an opinion as to one or the other style, but lean towards "Won't Fix". Let me know if you feel strongly otherwise!

PR #260, opened 2017-02-16
Consolidate comments about multi-view validation.

jsha:
Previously these were spread among the various challenges. We can simplify by
putting them in one place.

Replaces #239.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -2102,9 +2102,6 @@ using these steps:
    dNSName entries of SAN A and SAN B and no other entries.
    The comparison MUST be insensitive to case and ordering of names.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -2639,6 +2633,10 @@ particular case.  For identifiers where the server already has some public key
 associated with the domain this attack can be prevented by requiring the client
 to prove control of the corresponding private key.
 
+It is RECOMMENDED that the server perform DNS queries and make HTTP and TLS
+connections from various network perspectives, in order to make MitM attacks
```
rolandshoemaker: Does it make sense to reference this is talking about validations? I.e. `DNS queries and make HTTP and TLS connections use for validation`

PR #261, opened 2017-02-17
Switch more dictionary/key references with object/field

rolandshoemaker:
Follow-up to #248, catches a few more references I initially missed.


draft-ietf-acme-acme.md
```diff
@@ -843,7 +843,7 @@ the authorization expires.
 challenges (required, array):
 : The challenges that the client can fulfill in order to prove possession of the
 identifier (for pending authorizations).  For final authorizations, the
-challenges that were used.  Each array entry is a dictionary with parameters
+challenges that were used.  Each array entry is a object with parameters
```
jsha: *an object

rolandshoemaker:
I _think_ 'field name' makes more sense... but I'm not really wed to either.

PR #264, opened 2017-02-17
Update errors and add an error field to orders.

jsha:
Removes dnssec error type for a broader "dns" error type. It's very hard to split out dnssec errors from regular dns problems, because recursive resolvers can only return SERVFAIL. The new dns error type also subsumes "unknownHost."

Introduces a new "incorrectResponse" error type for cases where connection and request were both successful, but the content didn't match.

This also puts the validation-related errors at the bottom of the table so they can be viewed in context with each other.

Introduces an error field for orders. Orders can fail independent of authorizations because of last-minute CAA checking, failure to submit to CT logs, timeouts, etc.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -509,7 +509,7 @@ resources:
 |:----------------------|:-------------------------------------------------------------------|
 | caa                   | CAA records forbid the CA from issuing                             |
 | connection            | The server could not connect to validation target                  |
-| dns                   | There was a problem with a DNS query during validation             |
+| dns                   | There was a problem with a DNS query                               |
 | rateLimited           | The request exceeds a rate limit                                   |
 | serverInternal        | The server experienced an internal error                           |
 | incorrectResponse     | Response received didn't match the challenge's requirements        |
@@ -517,13 +517,12 @@ resources:
 
 Error types that may be commonly included in the "error" field of order
 resources:
-| unknownHost           | The server could not resolve a domain name                         |
-| unsupportedIdentifier | Identifier is not supported but may be in the future               |
-| userActionRequired    | The user visit the "instance" URL and take actions specified there |
 
 | Type                  | Description                                                        |
 |:----------------------|:-------------------------------------------------------------------|
 | caa                   | CAA records forbid the CA from issuing                             |
+| dns                   | There was a problem with a DNS query                               |
+| unsupportedIdentifier | Identifier is not supported but may be in the future               |
 | serverInternal        | The server experienced an internal error                           |
 
 These lists are not exhaustive. The server MAY return errors whose "type" field is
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -479,28 +479,38 @@ server MAY return status code 405 (Method Not Allowed).
...
 | unauthorized          | The client lacks sufficient authorization                          |
-| unknownHost           | The server could not resolve a domain name                         |
 | unsupportedIdentifier | Identifier is not supported, but may be in future                  |
-| userActionRequired    | The user visit the "instance" URL and take actions specified there |
+| userActionRequired    | Visit the "instance" URL and take actions specified there          |
+
+Error types that may be commonly included in the "error" field of challenge
+resources:
+
+| Code                  | Description                                                        |
```
rolandshoemaker: `Type` instead of `Code`

draft-ietf-acme-acme.md
```diff
@@ -765,6 +783,12 @@ notAfter (optional, string):
 : The requested value of the notAfter field in the certificate, in the date
 format defined in {{!RFC3339}}.
 
+error (optional, object):
+: The error that occurred while the server was issuing the certificate, if any.
```
jsoref: I don't really like this "while".

And the Note at the end of this paragraph should hint at the problem.

Instead:
"An error that occurred after authorizations were validated but before the server was able to issue a certificate (including its attempt to issue the certificate)." or
"An error that occurred after authorizations were validated but that prevented the server from issuing the certificate."

This obviates the need for the note.

jsha: Agreed, updated.

draft-ietf-acme-acme.md
```diff
@@ -484,29 +484,49 @@ server MAY return status code 405 (Method Not Allowed).
...
+resources:
+
+| Type                  | Description                                                        |
+|:----------------------|:-------------------------------------------------------------------|
+| caa                   | CAA records forbid the CA from issuing                             |
+| dns                   | There was a problem with a DNS query                               |
+| unsupportedIdentifier | Identifier is not supported but may be in the future               |
+| serverInternal        | The server experienced an internal error                           |
+
+These lists are not exhaustive. The server MAY return errors whose "type" field is
```
bifurcation: This change is problematic.  You've taken a normative list of all error codes to be registered for this protocol (and which IANA is going to use to initialize a registry) and turned it into three tables of examples. Instead, this section should list all error codes, and the specific lists of examples should move to the corresponding sections.

draft-ietf-acme-acme.md
```diff
@@ -771,6 +791,10 @@ notAfter (optional, string):
 : The requested value of the notAfter field in the certificate, in the date
 format defined in {{!RFC3339}}.
 
+error (optional, object):
+: The error that occurred while processing the order, if any.
+This field is structured as a problem document {{!RFC7807}}.
```
bifurcation: This is fine with me.

PR #265, opened 2017-02-18
Various small edits

rolandshoemaker:
Also moves various scattered character encoding references to their own section.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -960,9 +960,10 @@ invalid or unsupported contact URL, then the server MUST return an error of type
 "invalidContact", with a description describing the error and what types of
 contact URL the server considers acceptable.
 
-The server populates the account object with the "key" field with the JWK public key
-used to verify the JWS.  The server returns this account object in a response with
-a 201 (Created) status code, with the account URI in a Location header field.
+The server creates an account and populates the account object with the "key"
+field containing the JWK public key used to verify the JWS.  The server returns
+this account object in a response with a 201 (Created) status code, with the
+account URI in a Location header field.
 
 If the server already has an account registered with the provided account key,
 then it MUST return a response with a 200 (OK) status code and provide the URI of
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1069,8 +1069,9 @@ field.  This can be used to an ACME account with an existing account in a
 non-ACME system, such as a CA customer database.
 
 To enable ACME account binding, a CA needs to provision the ACME client with a
-MAC key and a key identifier. The MAC key SHOULD be provided in base64url-encoded
-form, to maximize compatibility between provisioning systems and ACME clients.
+MAC key and a key identifier. The key identifier MUST be an ASCII string. The
+MAC key SHOULD be provided in base64url-encoded form, to maximize compatibility
+between provisioning systems and ACME clients.
 
 The ACME client then computes a binding JWS to indicate the external account's
 approval of the ACME account key.  The payload of this JWS is the account key
@@ -1932,6 +1933,8 @@ for the domain in question.
 
 The path at which the resource is provisioned is comprised of the fixed prefix
 ".well-known/acme-challenge/", followed by the "token" value in the challenge.
+The value of the resource MUST be the ASCII representation of the key
+authorization.
 
 ~~~~~~~~~~
 GET .well-known/acme-challenge/evaGxfADs6pSRb2LAv9IZf17
@@ -2100,7 +2103,7 @@ Otherwise, the validation fails.
 ## DNS
 
 When the identifier being validated is a domain name, the client can prove
-control of that domain by provisioning a TXT record containing a designated
+control of that domain by provisioning a TXT resource record containing a designated
 value for a specific validation domain name.
 
 type (required, string):
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -298,11 +298,10 @@ character set. Trailing '=' characters MUST be stripped.
 ## Request Authentication
 
 All ACME requests with a non-empty body MUST encapsulate their payload
-in a JSON Web Signature (JWS) {{!RFC7515}} object, signed (in most
-cases) using the account's private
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1597,14 +1596,12 @@ Link: <https://example.com/acme/some-directory>;rel="directory"
...
-required information in the elements of the "challenges" dictionary.  (This is
-also the stage where the client should perform any actions required by the
-challenge.)
+required information in the elements of the "challenges" dictionary.
 
 The client sends these updates back to the server in the form of a JSON object
 with the response fields required by the challenge type, carried in a POST
-request to the challenge URI (not authorization URI).  This allows the client to
-send information only for challenges it is responding to.
+request to the challenge URI (not authorization URI) once it is ready for
```
jsha: This is passive voice. Instead "once it is ready for the server to attempt validation."

draft-ietf-acme-acme.md
```diff
@@ -1765,15 +1762,14 @@ Revocation requests are different from other ACME request in that they can be
 signed either with an account key pair or the key pair in the certificate.
 Before revoking a certificate, the server MUST verify that the key used to sign
 the request is authorized to revoke the certificate.  The server SHOULD consider
-at least the following accounts authorized for a given certificate:
+at least the following keys authorized to revoke a given certificate:
```
jsha: Please leave this section as-is. We recently went through one round of revisions to bring it to the state it's in today. In particular, I want to be clear that accounts are authorized as opposed to keys, with the one exception of the revoke-by-private-key.

draft-ietf-acme-acme.md
```diff
@@ -1067,9 +1069,8 @@ field.  This can be used to an ACME account with an existing account in a
 non-ACME system, such as a CA customer database.
 
 To enable ACME account binding, a CA needs to provision the ACME client with a
-MAC key and a key identifier.  The key identifier MUST be an ASCII string.  The
```
jsha: "The key identifier MUST be an ASCII string" is normative language. Let's leave it in for this "small edits" PR.

draft-ietf-acme-acme.md
```diff
@@ -1944,11 +1932,13 @@ for the domain in question.
 
 The path at which the resource is provisioned is comprised of the fixed prefix
 ".well-known/acme-challenge/", followed by the "token" value in the challenge.
-The value of the resource MUST be the ASCII representation of the key
```
jsha: This is also a change to normative language, so let's keep it in for now. Also, I'm not sure why this can be deleted. Is it redundant with somewhere else?

draft-ietf-acme-acme.md
```diff
@@ -2111,9 +2100,8 @@ Otherwise, the validation fails.
 ## DNS
 
 When the identifier being validated is a domain name, the client can prove
-control of that domain by provisioning a resource record under it.   The DNS
-challenge requires the client to provision a TXT record containing a designated
-value under a specific validation domain name.
+control of that domain by provisioning a TXT record containing a designated
```
jsha: s/record/resource record/.

draft-ietf-acme-acme.md
rolandshoemaker: This should be covered by the new normative language in the top-level `Character Encoding` section since ASCII is a subset of UTF-8 (technically).

draft-ietf-acme-acme.md
rolandshoemaker: This seemed redundant to me as DNS only (currently) supports the ASCII character set and therefore seems to repeat guidance from numerous other DNS RFCs.

jsha: You might be thinking of another part, since this section is talking about the HTTP challenge at `.well-known/acme-challenge/`.

rolandshoemaker: Oh whoops, yup got mixed up about which part of the diff I was looking at. This should also be covered by the `validation servers` reference in the new top-level `Character Encoding` section.

jsha: Okay, but this sentence is not just about "ASCII representation." It's also about what the response content should be.

rolandshoemaker: I would argue that is covered in the paragraph above by

> The client then provisions the key authorization as a resource on the HTTP server for the domain in question.

rolandshoemaker: Although I guess that doesn't have the normative language... I'd be happy to keep this line (minus the ASCII reference) and delete that line from the preceding parahraph as a compromise?

jsha: Let's just leave untouched the two bits that use normative language, and break out those fixes into their own PR so people can look a little more closely at them. This PR is pretty big.

draft-ietf-acme-acme.md
```diff
@@ -955,16 +960,14 @@ invalid or unsupported contact URL, then the server MUST return an error of type
 "invalidContact", with a description describing the error and what types of
 contact URL the server considers acceptable.
 
-The server creates an account object with the included contact information.  The
-"key" element of the account is set to the public key used to verify the JWS
-(i.e., the "jwk" element of the JWS header).  The server returns this account
-object in a 201 (Created) response, with the account URI in a Location header
-field.
+The server populates the account object with the "key" field with the JWK public key
```
jsha: Can you explain the intended change in this section?

rolandshoemaker: This was meant to simplify the second line that was removed (`The "key" element of the account is set to ...`) and remove the ambiguity of what is meant by `creates an account object` when the server is essentially just setting a single field in the user-provided object and returning it.

jsha: I think the old wording "The server creates an account object" is meant to indicate that the server is creating something in its storage, and assigning it a URL. I think that probably deserves to stay in. Otherwise it sounds like the server is just echoing back the account object created by the client, and not saving it.

rolandshoemaker: How would you feel about `The server creates an account and populates the account object with ...`, I think it gets across that the server has to do something but keeps it a bit simpler.

jsha: Perfect.

PR #266, opened 2017-02-18
Remove "key" from account object.

jsha:
The client knows what its public key is, because it has to have access to the private key in order to sign new-account requests or updates to the account resource. So this field serves no purpose.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -673,11 +673,6 @@ Content-Type: application/json
 An ACME account resource represents a set of metadata associated with an account.
 Account resources have the following structure:
 
-key (required, dictionary):
-: The public key of the account's key pair, encoded as a JSON Web Key object
-{{!RFC7517}}. The client may not directly update this field, but must use the
-key-change resource instead.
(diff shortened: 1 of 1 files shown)


bifurcation:
Has this been discussed on list?  I'm concerned that we're removing information that could be valuable for debugging client issues.  There's no harm in keeping this; the server has to store this information anyway.

jsha:
Yep, this was discussed on-list.

I can say that as a client developer, I've never found this field useful. Clients need to be able to successfully serialize a jwk in order to speak the protocol.

Also, I have no objection to servers including this as an extra field, but it seems wrong to normatively require it when it serves no purpose.

bifurcation:
For the record [this is the thread](https://mailarchive.ietf.org/arch/msg/acme/g7Zx2EmaaYCFK-u1s5R98xsbUh0/?qid=1f534919cd2c482b8cc1efdb93e2c863) where it was available for discussion.

For me, the observation that you need the private key to get an account object is dispositive.  There's not even debugging utility here.  So I'm OK to merge.

PR #272, opened 2017-03-10
Use correct challenge names in the IANA list

rolandshoemaker:


jsha:
This is worth discussion on the list. Do we consider the current, numbered, names a draft-only thing? Do we want the final RFC to use non-numbered challenge names?

bifurcation:
If we are going to do this, we need to update the following paragraph:

```
[[ Editor’s Note: In pre-RFC versions of this specification, challenges are labeled by type, and with the version of the draft in which they were introduced. For example, if an HTTP challenge were introduced in version -03 and a breaking change made in version -05, then there would be a challenge labeled “http-03” and one labeled “http-05” – but not one labeled “http-04”, since challenge in version -04 was compatible with one in version -03. ]]
```

Personally, I would prefer dropping the version number (i.e., not merging this PR), since if all goes well, future client development can be based on the RFC vs. the drafts.  Clients will need to change which token they use, but that's hardly the biggest change between what's deployed today and what's in this spec.

rolandshoemaker:
Will update to remove the editor's note.

rolandshoemaker:
Could possibly use some additional language about the name + version scheme we are using but I couldn't come up with anything that really worked.

PR #276, opened 2017-03-12
Register application/pem-certificate-chain.

jsha:
And update certificate section to use it. Also describe Link: rel="up" behavior
for pkix-cert.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -329,9 +329,10 @@ key, there MUST be a "jwk" field.
 For all other requests, there MUST be a "kid" field. This field must
 contain the account URI received by POSTing to the new-account resource.
 
-Note that authentication via signed JWS request bodies implies that GET
-requests are not authenticated.  Servers MUST NOT respond to GET requests
-for resources that might be considered sensitive.
+Note that authentication via signed JWS request bodies implies that GET requests
+are not authenticated.  Servers MUST NOT respond to GET requests for resources
+that might be considered sensitive.  Account resources are the only sensitive
+resources defined in this specification.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -2245,6 +2247,39 @@ identifier possession are determined by the server's local policy.
...
+
+Optional parameters: None
+
+Encoding considerations: None
+
+Security considerations: Carries a cryptographic certificate
+
+Interoperability considerations: None
+
+Published specification: draft-ietf-acme-acme
```
bifurcation: This should contain a note to the RFC editor to replace the value of the field with `RFC XXXX`, where `XXXX` is the assigned RFC number for this specification.

draft-ietf-acme-acme.md
```diff
@@ -1474,16 +1474,10 @@ described in {{identifier-authorization}} to complete the authorization process.
 To download the issued certificate, the client simply sends a GET request to the
 certificate URL.
 
-The default format of the certificate is PEM (application/x-pem-file) as
-specified by {{!RFC7468}}. This format should contain the end-entity certificate
-first, followed by any intermediate certificates that are needed to build a path
-to a trusted root. Servers SHOULD NOT include self-signed trust anchors. The
-client may request other formats by including an Accept header in its request.
-For example, the client could use the media type `application/pkix-cert`
-{{!RFC2585}} to request the end-entity certificate in DER format.
```
bifurcation: Was the idea of removing content negotiation discussed on the list?  I'm not seeing it.

jsha: I didn't remove content negotiation, just moved it lower in the section.

PR #281, opened 2017-03-20
Replaces HTTP(S) URIs with URLs

sophie-h:
This PR addresses my following comment on the mailing list.

> 4.
>   Is it really an account URI or rather an account URL? Some fields
>   changed from "uri" to "url". I guess, all occurrences of URIs should
>   be URLs. The only exception that I see is the "type" field in the
>   "Problem Details" responses.

Things that remain URIs

- "type" field in problem details
- "contact" fields in account objects

Needs clarification

- Should "terms-of-service" remain a URI? Seems unlikely that they are published as book.
- There is a part
  > server to query an arbitrary URI

  I think that should be a URL too.
-  > provide an integrity mechanism, which protects against an intermediary
   > changing the request URL to another ACME URL

   Right know I think the wording "ACME URL" makes sense. But it could be clarified why this is a *ACME* URL.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -319,7 +319,7 @@ JWS objects sent in ACME requests MUST meet the following additional criteria:
   * "jwk" (only for requests to new-account and revoke-cert resources)
   * "kid" (for all other requests)
   * "nonce" (defined in {{replay-protection}} below)
-  * "url" (defined in {{request-uri-integrity}} below)
+  * "url" (defined in {{request-url-integrity}} below)
 
 The "jwk" and "kid" fields are mutually exclusive. Servers MUST reject requests
 that contain both.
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -355,14 +355,14 @@ network terminate TLS connections from clients so that it can inspect client
 requests for denial-of-service protection.
 
 These intermediaries can also change values in the request that are not signed
-in the HTTPS request, e.g., the request URI and headers.  ACME uses JWS to
+in the HTTPS request, e.g., the request URL and headers.  ACME uses JWS to
 provide an integrity mechanism, which protects against an intermediary
-changing the request URI to another ACME URI.
+changing the request URL to another ACME URL.
```
cpu: Instead of "to another ACME URL" what about "to the URL of another ACME resource"? I think that might help clarify this sentence.

sophie-h:
Maybe I will just create separate PRs for the parts that are not covered yet. Then, they can be discussed separately.

PR #284, opened 2017-03-22
Rewords token entropy justification.

cpu:
Per feedback from Zach Shepherd[0] on the mailing list the justification
for the 128 bits of required token entropy from section 8.2 is at odds
with the actual rationale (as I understand it).

This commit updates the justification for the entropy requirement to not
mention being secret from an attacker, but instead not predictable by
a naive challenge response server that would complete challenges
without having been a party to the authz/chal creation.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1866,7 +1866,7 @@ type (required, string):
 
 token (required, string):
 : A random value that uniquely identifies the challenge.  This value MUST have
-at least 128 bits of entropy, in order to prevent an attacker from guessing it.
+at least 128 bits of entropy in order to prevent fully automatic challenge response without knowing the token from challenge creation.
 It MUST NOT contain any characters outside the base64url alphabet, including
 padding characters ("=").
 
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -2761,6 +2761,17 @@ server is in a default virtual host configuration.  Conversely, if the TLS
 server returns an unrecognized_name alert, then this is an indication that the
 server is not in a default virtual host configuration.
 
+## Token Entropy
+
+The http-01, tls-sni-02 and dns-01 validation methods mandate the usage of
+a random token value to uniquely identify the challenge. The value of the token
+is required to contain at least 182 bits of entropy for the following security
+properties. First, the ACME client should not be able to influence the ACME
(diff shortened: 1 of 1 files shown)


zjs:
It seems like whatever change is made should be applied to sections 8.2, 8.3, and 8.4.

jsha:
Based on zjs' feedback, I think it makes more sense for the individual token fields to just say "MUST have 128 bits", and explain why in one place in the Security Considerations section.

cpu:
@zjs @jsha Updated. Thanks for the feedback!

draft-ietf-acme-acme.md
```diff
@@ -2761,6 +2761,17 @@ server is in a default virtual host configuration.  Conversely, if the TLS
 server returns an unrecognized_name alert, then this is an indication that the
 server is not in a default virtual host configuration.
 
+## Token Entropy
+
+The http-01, tls-sni-02 and dns-01 validation methods mandate the usage of
+a random token value to uniquely identify the challenge. The value of the token
+is required to contain at least 182 bits of entropy for the following security
```
bifurcation: "128 bits"

cpu: Thanks for catching that :mag: :keyboard:. Fixed in https://github.com/ietf-wg-acme/acme/pull/284/commits/39d777e11e95841c569acd8465f3b1903f83f0b7

draft-ietf-acme-acme.md
```diff
@@ -1867,7 +1867,7 @@ type (required, string):
 
 token (required, string):
 : A random value that uniquely identifies the challenge.  This value MUST have
-at least 128 bits of entropy, in order to prevent an attacker from guessing it.
+at least 128 bits of entropy.
```
zjs: This same change should be made on [line 1972](https://github.com/ietf-wg-acme/acme/pull/284/files#diff-8430e2aa241beb4ac49b252db20d4ee8R1972) and [line 2074](https://github.com/ietf-wg-acme/acme/pull/284/files#diff-8430e2aa241beb4ac49b252db20d4ee8R2074).

cpu: @zjs Good catch - thanks! Updated in https://github.com/ietf-wg-acme/acme/pull/284/commits/1eece0d781c62e5f047ea292f7838851f9d4586d

PR #288, opened 2017-03-25
Fixes and clarifications

ohemorange:
Various minor clarifications and fixes

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -161,8 +161,8 @@ The two main roles in ACME are “client” and “server”.  The ACME client u
 protocol to request certificate management actions, such as issuance or
 revocation.  An ACME client may run on a web server, mail server, or some other
 server system which requires valid TLS certificates.  Or, it may run on a separate
-server that does not consume the certificate, but is authorized to respond to the
-identifier validation challenge.  The ACME server runs at a certification authority,
+server that does not consume the certificate, but is authorized to respond to a
+CA-provided challenge.  The ACME server runs at a certification authority,
 and responds to client requests, performing the requested actions if the client is
 authorized.
 
@@ -1994,7 +1994,7 @@ the challenge.
 
 The certificate may be constructed arbitrarily, except that each certificate
 MUST have exactly two subjectAlternativeNames, SAN A and SAN B. Both MUST be
-dNSNames {{!RFC2585}}.
+dNSNames {{!RFC5280}}.
 
 SAN A MUST be constructed as follows: compute the SHA-256 digest [FIPS180-4] of
 the challenge token and encode it in lowercase hexadecimal form.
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -124,8 +124,8 @@ no out-of-band human intervention.
 
 Prior to ACME, when deploying an HTTPS server, an operator typically gets a
 prompt to generate a self-signed certificate.  If the operator were instead
-deploying an ACME-compatible web server, the experience would be something like
-this:
+deploying an HTTPS server using an ACME client, the experience would
```
jsha: I think we should be agnostic to "HTTPS server + ACME client" (like Certbot) vs "HTTPS server with built-in ACME client" (like Caddy), especially since we'd like more HTTPS servers to build in automatic ACME support.

draft-ietf-acme-acme.md
```diff
@@ -159,10 +159,12 @@ document are to be interpreted as described in RFC 2119 {{!RFC2119}}.
...
 The two main roles in ACME are “client” and “server”.  The ACME client uses the
 protocol to request certificate management actions, such as issuance or
-revocation.  An ACME client therefore typically runs on a web server, mail
-server, or some other server system which requires valid TLS certificates.  The
-ACME server runs at a certification authority, and responds to client requests,
-performing the requested actions if the client is authorized.
+revocation.  An ACME client may run on a web server, mail server, or some other
+server system which requires valid TLS certificates.  Or, it may run on a separate
+server that does not consume the certificate, but is authorized to respond to the
+identifier validation challenge.  The ACME server runs at a certification authority,
```
jsha: We don't define "identifier validation challenge" above here, but we do talk about "a CA-provided challenge," so let's reuse that terminology here.

draft-ietf-acme-acme.md
```diff
@@ -1992,7 +1994,7 @@ the challenge.
 
 The certificate may be constructed arbitrarily, except that each certificate
 MUST have exactly two subjectAlternativeNames, SAN A and SAN B. Both MUST be
-dNSNames.
+dNSNames {{!RFC2585}}.
```
jsha: I think RFC 5280 is a better reference here.

draft-ietf-acme-acme.md
ohemorange: Great point! How about "If the operator were instead deploying an HTTPS server using ACME", and exploit ambiguity in the English language?

draft-ietf-acme-acme.md
ohemorange: oh definitely

draft-ietf-acme-acme.md
jsha: Sounds good to me!

PR #289, opened 2017-03-27
Make Replay-Nonce uniqueness a MUST

jsha:
And clarify "high probability."


bifurcation:
While we're clarifying, I might add an encrypted counter another option that is guaranteed unique.

jsha:
I don't think we need to get into that level of prescribing implementation detail; once you mention it at all it immediately sounds half-specified. As a reader, I would want to know "what algorithm? shared across how many frontends? how does key management and rotation work?" Better to just say it should be unique (with high probability) and leave it to the implementer.

PR #290, opened 2017-03-31
Add CAA text (per IETF98)

richsalz:


draft-ietf-acme-acme.md
```diff
@@ -2726,6 +2726,10 @@ within their trusted network and use these resolvers both for both CAA record
 lookups and all record lookups in furtherance of a challenge scheme (A, AAAA,
 TXT, etc.).
 
+Further, an ACME-based CA can use the Certification Authority Authorization
```
yaronf: Do we have consensus on using the non-normative "can" here, instead of e.g. SHOULD? CAA is very much security-relevant.

draft-ietf-acme-acme.md
```diff
@@ -2726,6 +2726,10 @@ within their trusted network and use these resolvers both for both CAA record
...
+record {{!RFC6844}} to prevent it from being misdirected and generate
```
jsha: *and generating

draft-ietf-acme-acme.md
richsalz: okay, used the "It is RECOMMENDED" language used elsewhere in this section.

draft-ietf-acme-acme.md
richsalz: Actually, "misirected and thereby generate" ...

richsalz:
Update commit pushed.

jsha:
I meant that there's a tense mismatch between "being misdirected" and "generate."

> being misdirected and thereby generate an unauthorized issuance.

richsalz:
No there isn't.  "From being misdirected and then generate (as a result of that)..."  They occur at two different times :)
But feel free to wordsmith however you want.

bifurcation:
Closing in favor of #293

PR #291, opened 2017-04-04
Adds "unsupportedContact" error, makes "mailto:" MTI.

cpu:
Firstly, this PR updates the table of errors in Section 6.6
"Errors" to add an "unsupportedContact" error type in addition to
"invalidContact". The former is returned when the contact URI protocol
is unsupported and the latter when the protocol is supported but the
provided contact is invalid.

Secondly, this PR updates Section 7.3 "Account Creation" to both
mention the new "unsupportedContact" error type and to make support of
the "mailto:" protocol a MUST.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -940,7 +940,7 @@ support an extension field.
 
 The server SHOULD validate that the contact URIs in the "contact" field are
 valid and supported by the server. If the server validates contact URIs it MUST
-support the mailto scheme. If the server rejects a contact URI for using an
+support the "mailto" scheme. If the server rejects a contact URI for using an
 unsupported scheme it MUST return an error of type "unsupportedContact", with
 a description describing the error and what types of contact URIs the server
 considers acceptable. If the server rejects a contact URI for using a supported
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -939,12 +939,13 @@ resulting account object.  This allows clients to detect when servers do not
 support an extension field.
 
 The server SHOULD validate that the contact URIs in the "contact" field are
-valid and supported by the server. If the client provides the server with
-a contact URI using an unsupported protocol it MUST return an error of type
-"unsupportedContact", with a description describing the error and what types of
-contact URIs the server considers acceptable. The server MUST support the mailto
-scheme for contact URIs. If the client provides the server with an invalid
-contact URI, then the server MUST return an error of type "invalidContact".
+valid and supported by the server. If the server validates contact URIs it MUST
+support the mailto scheme. If the server rejects a contact URI for using an
+unsupported scheme it MUST return an error of type "unsupportedContact", with
+a description describing the error and what types of contact URIs the server
+considers acceptable. If the server rejects a contact URI for using a supported
+scheme but an invalid value then the server MUST return an error of type
+"invalidContact".
 
 The server creates an account and stores the public key used to verify the
 JWS (i.e., the "jwk" element of the JWS header) to authenticate future requests
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -938,13 +938,13 @@ not recognize.  In particular, it MUST NOT reflect unrecognized fields in the
 resulting account object.  This allows clients to detect when servers do not
 support an extension field.
 
-The server SHOULD validate that the contact URLs in the "contact" field are
+The server SHOULD validate that the contact URIs in the "contact" field are
 valid and supported by the server. If the client provides the server with
-a contact URL using an unsupported protocol it MUST return an error of type
-"unsupportedContact". The server MUST support the "mailto:" protocol. If the
-client provides the server with an invalid contact URL, then the server MUST
-return an error of type "invalidContact", with a description describing the
-error and what types of contact URL the server considers acceptable.
+a contact URI using an unsupported protocol it MUST return an error of type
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -475,6 +475,7 @@ in the "type" field (within the "urn:ietf:params:acme:error:" namespace):
 | badNonce              | The client sent an unacceptable anti-replay nonce                              |
 | badSignatureAlgorithm | The JWS was signed with an algorithm the server does not support               |
 | invalidContact        | The contact URI for an account was invalid                                     |
+| unsupportedContact    | The contact URI for an account used an unsupported protocol scheme             |
```
zjs: Should this (and `invalidContact`) read "A contact URI ..." ("the" seems like the wrong article, since an account may have more than one).

draft-ietf-acme-acme.md
```diff
@@ -938,10 +939,12 @@ resulting account object.  This allows clients to detect when servers do not
 support an extension field.
 
 The server SHOULD validate that the contact URLs in the "contact" field are
-valid and supported by the server.  If the client provides the server with an
-invalid or unsupported contact URL, then the server MUST return an error of type
-"invalidContact", with a description describing the error and what types of
-contact URL the server considers acceptable.
+valid and supported by the server. If the client provides the server with
+a contact URL using an unsupported protocol it MUST return an error of type
+"unsupportedContact". The server MUST support the "mailto:" protocol. If the
```
zjs: > "mailto:" protocol

Based on [RFC 6068](https://tools.ietf.org/html/rfc6068) and [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.1), it seems like we should refer to this as "mailto scheme" (no quotes, no colon, using scheme instead of protocol).

> The server MUST support the "mailto:" protocol.

I think some context is warranted here; we don't actually expect ACME servers to support mailto (i.e., for the server to implement the mailto protocol), we expect ACME servers to support the mailto scheme _for contact URIs_.

draft-ietf-acme-acme.md
```diff
@@ -938,10 +939,12 @@ resulting account object.  This allows clients to detect when servers do not
...
+client provides the server with an invalid contact URL, then the server MUST
+return an error of type "invalidContact", with a description describing the
+error and what types of contact URL the server considers acceptable.
```
zjs: > what types of contact URL the server considers acceptable

Should this clause be moved to `unsupportedContact`?

draft-ietf-acme-acme.md
cpu: Good point. I will update the wording for `invalidContact` and `unsupportedContact`.

draft-ietf-acme-acme.md
cpu: > Based on RFC 6068 and RFC 3986, it seems like we should refer to this as "mailto scheme" (no quotes, no colon, using scheme instead of protocol).

WFM. Thanks!

> I think some context is warranted here; we don't actually expect ACME servers to support mailto (i.e., for the server to implement the mailto protocol), we expect ACME servers to support the mailto scheme for contact URIs.

+1 - I'll clarify that.

cpu:
Thanks for the feedback @zjs - updated.

draft-ietf-acme-acme.md
```diff
@@ -937,11 +938,13 @@ not recognize.  In particular, it MUST NOT reflect unrecognized fields in the
...
-invalid or unsupported contact URL, then the server MUST return an error of type
-"invalidContact", with a description describing the error and what types of
-contact URL the server considers acceptable.
+The server SHOULD validate that the contact URIs in the "contact" field are
+valid and supported by the server. If the client provides the server with
+a contact URI using an unsupported protocol it MUST return an error of type
+"unsupportedContact", with a description describing the error and what types of
+contact URIs the server considers acceptable. The server MUST support the mailto
+scheme for contact URIs. If the client provides the server with an invalid
+contact URI, then the server MUST return an error of type "invalidContact".
```
zjs: The changes look good, but in re-reading this I noticed a pre-existing issue.

It seems like validation described in the first sentence ("The server SHOULD validate that the contact URIs in the "contact" field are valid ...) would be required to adhere to the MUST from the last sentence ("If the client provides the server with an invalid contact URI, then the server MUST ..."). 

It seems like there are a few ways to address this. Two that come to mind would be:
(a) make the first sentence of this paragraph a MUST as well
(b) further qualify the two MUSTs so that they are mandatory only if the server implements validation (e.g., "If the server validates the contact URI and the client provides the server with an invalid contact URI, then ...")

cpu: I reworded this in https://github.com/ietf-wg-acme/acme/pull/291/commits/44762759546b528843899bedc638426a20556677 and I think it makes it clearer that the errors are for servers that choose to validate.

cpu: Thanks!

daothithuy:
Dao Thi Thuy was born on 1988 August Hamlet of Dinh An Commune, Cho Moi District, An Giang Province Hello everyone i used to understand my second leg is the same but it is only 28 and 29 is that tree Comedy web site are the same want to use 2 web pages to use word of communication cone width as I want my cat to be real name for my testimony and Fresh Country

uhhhh2:
Should the "unsupportedContact" error include a separate machine-readable field with the types of contacts acceptable?

bifurcation:
@uhhhh2 Let's discuss that in a follow-on issue / PR.

PR #292, opened 2017-04-09
Allow challenge retries

hlandau:
As discussed on list, challenges should be retriable.


jsha:
Closing in preference for #297. Thanks for proposing!

PR #304, opened 2017-04-27
Change where polling happens to be order.

### LINKED ISSUE #237: Recommend polling order rather than authz
Per [comment](https://github.com/ietf-wg-acme/acme/pull/232/files#r97096663) in #232.

jsha:
Fixes #237


jsha:
Per minutes, Hugo gave feedback based on retriable challenges that this should stay with polling authzs. Closing.

PR #307, opened 2017-04-27
Forbid rolling over to a duplicate key

bifurcation:


draft-ietf-acme-acme.md
```diff
@@ -1230,11 +1230,14 @@ addition to the typical JWS validation:
    the account matching the old key
 8. Check that the "newKey" field of the key-change object also verifies the
     inner JWS.
+9. Check that there is no current account whose account key is the same as the
```
zjs: It doesn't seem like "current account" is a term used elsewhere in the document. Could this be rephrased to avoid introducing a new term? (Presumably this is referring to accounts with a status of "active"?)

PR #308, opened 2017-04-27
Clarify behavior of only-return-existing with no existing

bifurcation:


jsha:
Is 200 really correct? This seems like a 404 maybe? Also we should specify the body (empty).

uhhhh2:
Would a problem document be appropriate (with either a pre-existing ACME error type or a new one like "accountDoesNotExist")?

martinthomson:
404 talks about the effective request URI, so that's not going to work.

I think that a 400 would be better on the basis that the request checked, but failed (@uhhhh2's suggestion is a fine enhancement).  The reason you make this request is to check; and having a 4xx series code is easier to test than the absence of a header field.

bifurcation:
Thanks @martinthomson and @uhhhh2, implemented this in the latest commit.

draft-ietf-acme-acme.md
```diff
@@ -1007,8 +1008,8 @@ account to be created if one does not already exist, then it SHOULD do so by
 sending a POST request to the new-account URL with a JWS whose payload has an
 "only-return-existing" field set to "true" ({"only-return-existing": true}).
 If a client sends such a request and an account does not exist, then the server
-MUST return a response with a 200 (OK) status code and no Location header
-field.
+MUST return am error response with status code 400 (Bad Request) and type
```
uhhhh2: "an error"

PR #312, opened 2017-05-11
Protect HTTP validation from stateless clients

zjs:
As discussed on the mailing list[1], a stateless client can be written
for http-01 because the token is provided by the server as a part of
the validation request.

This change introduces http-07, which uses a SHA-256 digest of the token
in the request instead of the token itself, as suggested by Logan Widick.

Because the expected request body is unchanged between http-01 and
http-07, clients and servers could trivially support both during a
transition period.

Note: I think it may make more sense to call this http-02, especially if
      the version numbers in challanges are not going to be removed in
      the final version of the RFC, but this change follows the  current
      versioning strategy as described in the Editor's Note at the end
      of Section 8.

1 - https://www.ietf.org/mail-archive/web/acme/current/threads.html#01941

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -865,7 +865,7 @@ name validation.
   "challenges": [
     {
       "url": "https://example.com/authz/1234/0",
-      "type": "http-01",
+      "type": "http-07",
       "status": "valid",
       "token": "DGyRejmCefe7v4NfDGDKfA"
       "validated": "2014-12-01T12:05:00Z",
@@ -1568,7 +1568,7 @@ Link: <https://example.com/acme/some-directory>;rel="index"
 
   "challenges": [
     {
-      "type": "http-01",
+      "type": "http-07",
       "url": "https://example.com/authz/1234/0",
       "token": "DGyRejmCefe7v4NfDGDKfA"
     },
@@ -1598,7 +1598,7 @@ with the response fields required by the challenge type, carried in a POST
 request to the challenge URL (not authorization URL) once it is ready for
 the server to attempt validation.
 
-For example, if the client were to respond to the "http-01" challenge in the
+For example, if the client were to respond to the "http-07" challenge in the
 above authorization, it would send the following request:
 
 ~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1950,16 +1950,16 @@ client then provisions the key authorization as a resource on the HTTP server
 for the domain in question.
 
 The path at which the resource is provisioned is comprised of the fixed prefix
-".well-known/acme-challenge/", followed by the "token" value in the challenge.
-The value of the resource MUST be the ASCII representation of the key
-authorization.
+".well-known/acme-challenge/", followed by the SHA-256 digest of the "token"
+value in the challenge. The value of the resource MUST be the ASCII
```
bifurcation: You need to specify how this is encoded.  I would be inclined to use the same base64url encoding we use elsewhere, if only to avoid ambiguity about upper case / lower case hex.

Why not just digest the entire key authorization?  Then you're basically just requiring the client to give you a hash pre-image.

draft-ietf-acme-acme.md
```diff
@@ -2512,7 +2512,7 @@ Initial Contents
 
 | Label      | Identifier Type | Reference |
 |:-----------|:----------------|:----------|
-| http-01    | dns             | RFC XXXX  |
+| http-07    | dns             | RFC XXXX  |
```
bifurcation: I can agree with using `http-02` for this.

We should probably keep `http-01` in here as well.

bifurcation:
@zjs I'm trying to get a new version out the door so we can go to IETF Last Call.  So a prompt update would be appreciated.  Or if you agree with the proposals above, just let me know and I'll be happy to make the changes myself.

draft-ietf-acme-acme.md
zjs: I think base64url encoding makes perfect sense.

A hash of the token (instead of the entire key authorization) was proposed for consistency with SAN A from `tls-sni-02`. This would result in a de facto pattern that the SHA256 of the token is always used in requests, which could be applied to future challenge types.

draft-ietf-acme-acme.md
zjs: It doesn't seem like the table still contains `tls-sni-01`. Is there some reason to include `http-01`, but not `tls-sni-01`?

PR #332, opened 2017-07-07
Generalize the validation methods registry

bifurcation:
It is useful in some contexts (e.g., [CAA](https://ietf-wg-acme.github.io/acme-caa/)) to have identifiers for non-ACME methods that CAs use to validate control of an identifier.  This PR updates the ACME Challenge Types registry to instead address Validation Methods more generally, with a column that indicates whether the method is an ACME method.


draft-ietf-acme-acme.md
```diff
@@ -2506,27 +2506,29 @@ Initial contents:
...
 
 Template:
 
-* Label: The value to be put in the "type" field of challenge objects using this
-  validation mechanism
-* Identifier Type: The type of identifier that this mechanism applies to
-* Reference: Where the challenge type is defined
+* Label: The identifier for this validation method
+* Identifier Type: The type of identifier that this method applies to
+* ACME: "Y" if the validation method corresponds to an ACME challenge type;
```
richsalz: perhaps add "as defined by an IETF WG" to the end of this clause

bifurcation: In the current spec, you don't need IETF consensus to add to the challenge types registry.  So this change would be raising the bar.

In any case, what I want this column to capture is not some process point, only whether someone has defined how to do this method with ACME.

richsalz:
Okay, I see your point.  Fine.

draft-ietf-acme-acme.md
```diff
@@ -2506,27 +2506,29 @@ Initial contents:
 \[\[ RFC EDITOR: Please replace XXXX above with the RFC number assigned to this
 document ]]
 
-### Challenge Types {#iana-challenge}
+### Validation Methods {#iana-challenge}
```
jsha: Do you want to change the `{#iana-challenge}` to `{#iana-validation}`?

PR #342, opened 2017-09-26
Remove proactive issuance & csr-first new-order.

cpu:
This commit implements a first go at the replacement of proactive issuance
and the "CSR first" new-order flow [proposed on the ACME mailing list](https://mailarchive.ietf.org/arch/msg/acme/DIjJEB06J5cFyuOlGPVcY2I51vg)

Mentions of proactive issuance are removed. The CSR field of the order
object is removed. A new "finalizeURL" field is added to the order
object. A new "identifiers" field is added to the order object. Issuance
process is updated to describe submitting identifiers in the new-order
request and POSTing a CSR to the order's finalizeURL.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -803,8 +803,11 @@ certificate (optional, string):
 }
 ~~~~~~~~~~
 
-Identifiers in a new-order request MAY contain wildcard identifiers, even if
-the resulting authorizations returned by the server do not.
+Identifiers in a new-order request MAY contain DNS type identifiers with
+wildcard values, even if the resulting authorizations returned by the server do
+not. DNS type identifiers for a wildcard value MUST contain only one '*'
+wildcard character comprising the entire leftmost DNS label. (E.g.
+"*.foo.example.com" not "foo.*.example.com" or "*.*.example.com").
 
 The elements of the "authorizations" and "identifiers" array are immutable once
 set.  The server MUST NOT change the contents either array after they are
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -784,7 +784,11 @@ certificate (optional, string):
   "status": "valid",
   "expires": "2015-03-01T14:09:00Z",
 
-  "csr": "jcRf4uXra7FGYW5ZMewvV...rhlnznwy8YbpMGqwidEXfE",
+  "identifiers": [
+    { "type": "dns", "value": "example.com" },
+    { "type": "dns", "value": "www.example.com" }
+  ],
+
   "notBefore": "2016-01-01T00:00:00Z",
   "notAfter": "2016-01-08T00:00:00Z",
 
@@ -792,6 +796,8 @@ certificate (optional, string):
     "https://example.com/acme/authz/1234",
     "https://example.com/acme/authz/2345"
   ],
+
+  "finalizeURL": "https://example.com/acme/acct/1/order/1/finalize",
 
   "certificate": "https://example.com/acme/cert/1234"
 }
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1372,7 +1372,7 @@ Location: https://example.com/acme/order/asdf
     "https://example.com/acme/authz/2345"
   ],
 
-  "finalizeURL": "https://example.com/acme/order/finalize/asdf"
+  "finalizeURL": "https://example.com/acme/order/asdf/finalize"
 }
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -784,14 +797,19 @@ certificate (optional, string):
...
   "certificate": "https://example.com/acme/cert/1234"
 }
 ~~~~~~~~~~
 
-The elements of the "authorizations" array are immutable once set.  The server
-MUST NOT change the contents of the "authorizations" array after it is created.
-If a client observes a change in the contents of the "authorizations" array,
-then it SHOULD consider the order invalid.
+Identifiers in a new-order request MAY contain wildcard identifiers, even if
+the resulting authorizations returned by the server do not.
```
martinthomson: I think that you need to be a little more precise about defining the wildcard identifier for each type of identifier (i.e., the leftmost label can be '*' for DNS)

cpu: @martinthomson Thanks for the feedback. I took a shot at clarifying with https://github.com/ietf-wg-acme/acme/pull/342/commits/58eec446e678fa5aec7cf35d44162617f0aa6427 Open to further refinement!

Should there be an RFC reference for the requirements placed on wildcard DNS identifiers? If so what's the best RFC to cite?

I thought that RFC5280 might specify the "only one wildcard character, only as the entire left-most DNS label" behaviour but it [explicitly doesn't address the semantics of wildcard SAN names](https://tools.ietf.org/html/rfc5280#section-4.2.1.6). RFC6125 has a section on [checking wildcard certificates](https://tools.ietf.org/html/rfc6125#section-6.4.3) but it seems a little loosey-goosey and the last MAY is (to my understanding) not something browsers do.

jsha: I've done some digging on this, and it's a mess. RFC 5280 says dNSName SANs must be in RFC 1034 "preferred name syntax." Preferred name syntax is letter-digit-hyphen, no "*" allowed. Inserting "*"'s became common practice (against 5280), as documented in RFC 6125.

[Section 7.2 of RFC 6125](https://tools.ietf.org/html/rfc6125#section-7.2) describes the situation for issuers, versus 6.4.3, which you link to and which describes the situation for relying parties:
> This document states that the wildcard character '*' SHOULD NOT be
> included in presented identifiers but MAY be checked by application
> clients (mainly for the sake of backward compatibility with deployed
> infrastructure).

The Baseline Requirements say:

> Domain Name: The label assigned to a node in the Domain Name System.
> Fully-Qualified Domain Name: A Domain Name that includes the labels of all superior nodes in the Internet Domain Name System.
> Wildcard Domain Name: A Domain Name consisting of a single asterisk character followed by a single full stop character (“*.”) followed by a Fully-Qualified Domain Name.
> 7.1.4.2.1 Subject Alternative Name Extension
> ...
> Wildcard FQDNs are permitted.

So, AFAICT, RFC5280 doesn't allow wildcards, then RFC6125 extra discourages them, but they are in widespread practical use. I think your current revision is very close. Here's my proposal to bring it closer to the BR definition:

> Any identifier of type "dns" in a new-order request MAY have a wildcard domain
> name as its value. A wildcard domain name consists of a single asterisk
> character followed by a single full stop character (“*.”) followed by a
> domain name as defined for use in the Subject Alternative Name Extension by RFC
> 5280.

cpu: > I've done some digging on this, and it's a mess. 

Classic.

> So, AFAICT, RFC5280 doesn't allow wildcards, then RFC6125 extra discourages them, but they are in widespread practical use. I think your current revision is very close. Here's my proposal to bring it closer to the BR definition

Thanks! That reads much better. I stole your wording for f451f2c and preserved the "Your wildcard shouldn't be in the returned authorizations" part by more closely matching the MUST NOT from 7.1.4.

richsalz:
A wildcard is not a dnsName name.  It happens to use the same encoding, perhaps unfortunately.

cpu:
> A wildcard is not a dnsName name.  It happens to use the same encoding, perhaps unfortunately.

@richsalz Can you point out which line of the diff you're referring to? I don't see where in f451f2c we call a wildcard a dnsName.

richsalz:
I was referring to the RFC’s.

cpu:
> I was referring to the RFC’s.

Aha. So no changes being requested for this PR :+1:

bifurcation:
There's been a bunch of list discussion on this, and a pretty clear consensus call at IETF 100.  Merging based on the chairs call of consensus.

PR #346, opened 2017-10-19
Further restrict mailto URLs

bifurcation:


jsha:
This doesn't address the other piece of feedback, that mailto URLs can contain multiple entries.

bifurcation:
If you don't allow `hfields`, then it's only possible to have one address.

bifurcation:
Oh, no, that's totally wrong, will update.

ivanloveit:
Oh no

PR #350, opened 2017-11-06
WIP: Enable servers to require authorization before order

bifurcation:
An alternative solution to #342

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1395,15 +1395,10 @@ client fails to complete the required actions before the "expires" time, then
 the server SHOULD change the status of the order to "invalid" and MAY
 delete the order resource.
 
-The server MUST begin the issuance process for the requested certificate and
-update the order resource with a URL for the certificate once the client has
-fulfilled the server's requirements.   If the client has already satisfied the
-server's requirements at the time of this request (e.g., by obtaining
-authorization for all of the identifiers in the certificate in previous
-transactions), then the server MUST proactively issue the requested certificate
-and provide a URL for it in the "certificate" field of the order.  The server
-MUST, however, still list the completed authorizations in the "authorizations"
-array.
+Once the client has fulfilled the server's requirements, the server MUST update
+the order resoruce with a URL for the certificate.  It SHOULD begin the issuance
+process at this point, but MAY postpone issuance until it receives a GET request
+for the certificate URL.
 
 Once the client believes it has fulfilled the server's requirements, it should
 send a GET request to the order resource to obtain its current state.  The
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1394,15 +1395,10 @@ client fails to complete the required actions before the "expires" time, then
...
-update the order resource with a URL for the certificate once the client has
-fulfilled the server's requirements.   If the client has already satisfied the
-server's requirements at the time of this request (e.g., by obtaining
-authorization for all of the identifiers in the certificate in previous
-transactions), then the server MUST proactively issue the requested certificate
-and provide a URL for it in the "certificate" field of the order.  The server
-MUST, however, still list the completed authorizations in the "authorizations"
-array.
+Once the client has fulfilled the server's requirements, the server MUST update
+the order resoruce with a URL for the certificate.  It SHOULD begin the issuance
```
cpu: typo: "resoruce"

draft-ietf-acme-acme.md
```diff
@@ -1394,15 +1395,10 @@ client fails to complete the required actions before the "expires" time, then
...
+process at this point, but MAY postpone issuance until it receives a GET request
+for the certificate URL.
```
cpu: Let's Encrypt currently constructs certificate URLs using the serial # of the certificate. It's not a big deal but I wanted to note that a MUST for updating the order with a certificate URL when the order is authorized and a MAY for postponing issuance until GET means CAs that delay issuance can't reference any certificate attributes in the URL since they have to construct it before issuance.

draft-ietf-acme-acme.md
```diff
@@ -1419,8 +1415,63 @@ status of the order will indicate what action the client should take:
...
+
+In some cases, a server may require a client to complete the authorizations
+necessary to issue a certificate before it will accept a new-order request.
+This allows the server to avoid caching a CSR before authorizations have been
+completed.  When such a server receives a new-order request without sufficient
+authorizations, it should return an error response with status code 401
+(Unauthorized) and error code "authorizationFirst".  Any error document with
+this error code MUST include a field "authorizations" containing an array of
+URLs for authorization objects that must be completed before the new-order
+request can succeed.
```
cpu: This seems awkward - are there any other places in the protocol where the caller is expected to pull a non-standard field out of an error document in order to learn how to proceed with a subsequent API operation?

bifurcation: Don't know if this counts as a subsequent API operation, but the [section on re-agreement to ToS](https://ietf-wg-acme.github.io/acme/draft-ietf-acme-acme.html#rfc.section.7.3.4) has the client direct the user to a URI from the "instance" field.

In any case, the inclusion of additional data in problem documents is [well defined](https://tools.ietf.org/html/rfc7807#section-3.2).

PR #354, opened 2017-11-20
Define sub-problems.

jsha:


draft-ietf-acme-acme.md
```diff
@@ -521,6 +521,53 @@ set to a URI other than those defined above.  Servers MUST NOT use the ACME URN
 namespace for errors other than the standard types.  Clients SHOULD display the
 "detail" field of all errors.
 
+### Sub-problems
+
+Sometimes a CA may need to return multiple errors to a single
```
cpu: nitpicky nit: "return multiple errors to a request" reads better in my mind as "return multiple errors in response to a request"

draft-ietf-acme-acme.md
```diff
@@ -521,6 +521,53 @@ set to a URI other than those defined above.  Servers MUST NOT use the ACME URN
...
+request. Additionally, the CA may need to attribute errors to specific
+identifiers.  For instance, a new-order request may contain multiple
+identifiers for which the CA cannot issue. In this situation, an ACME
+problem document MAY contain the "sub-problems" field, contains a JSON
```
cpu: "contains a JSON array of problem documents" -> "containing a JSON array of problem documents"

draft-ietf-acme-acme.md
```diff
@@ -521,6 +521,53 @@ set to a URI other than those defined above.  Servers MUST NOT use the ACME URN
...
+array of problem documents, each of which MAY contain an "identifier"
+field. If present, the "identifier" field MUST contain an ACME identifier
+({{iana-identifier}}). The "identifier" field MUST NOT be present at
+the top level in ACME problem documents. It can only be present in sub-problems.
+Sub-problems need not all have the same type, and do not need to match the top level type.
+
+ACME clients may choose to use the "identifier" field as a hint that
```
cpu: "may choose to use the "identifier" field" -> "may choose to use the "identifier field of a sub-problem" ? I think "identifier field" on its own is perhaps a shade ambiguous.

draft-ietf-acme-acme.md
```diff
@@ -521,6 +521,53 @@ set to a URI other than those defined above.  Servers MUST NOT use the ACME URN
...
+of those identifiers, the ACME client may choose to submit another order
+containing only the eight identifiers not listed in the problem document.
+
+~~~~~
+HTTP/1.1 403 Forbidden
+Content-Type: application/problem+json
+
+{
+    "type": "urn:ietf:params:acme:error:malformed",
+    "detail": "Some of the identifiers requested were rejected",
```
cpu: the other problem examples in the spec (the `usrActionRequired` example in ToS mismatch & the `unauthorized` problem for a missing authz in revocation) both include an `"instance"` field pointing at a docs URL (e.g. `http://example.com/doc/unauthorized`.

I think both the toplevel problem and the sub-problems should probably have an `instance` field as well.

jsha: "instance" is intended to refer to the specific instance of the problem. For userActionRequired, this makes sense because you'd have a unique token to indicate re-agreement for that user. For the unauthorized example, I think the usage of "instance" is incorrect. I'll remove in a separate PR.

draft-ietf-acme-acme.md
```diff
@@ -521,6 +521,53 @@ set to a URI other than those defined above.  Servers MUST NOT use the ACME URN
...
+    "sub-problems": [
+        {
+            "type": "urn:ietf:params:acme:error:malformed",
+            "value": "Invalid underscore in DNS name \"_example.com\"",
```
shred: Shouldn't this be a `"detail"` key instead of `"value"`?

PR #357, opened 2017-11-28
List feedback

### LINKED ISSUE #328: OOB-01 should take an empty object payload
The other challenges don't expect you to post a payload containing a type field, so it's inconsistent for OOB-01 to do so. It should just take an empty object "{}" as its payload.

### LINKED ISSUE #329: Define "response fields" more clearly
In "Responding to challenges," we use the term "response fields." These fields are defined by each challenge type, but are not described as "response fields" there. We should add a heading describing them as such in the challenge type definitions.

### LINKED ISSUE #330: Fix feedback from list
On the mailing list, Rifaat Shekh-Yusuf pointed out these issues:

* Section 7.3.5, first paragraph, second line:
  A "bind" word is missing between the words "to" and "an"

* Section 7.4.1,
  - second paragraph, second sentence: "case" should be "cases".
  - When the server builds the authorization object, the document is
    stating that the response would include "challenges" and "combinations".
    Remove the "combinations" as it is not being used.

* Section 7.5.1, the Request URIs in the examples:
  Should not this be /acme/authz/1234/0?

bifurcation:
Addresses Martin Stiemerling's TSV-ART review
Fixes #328 
Fixes #329 
Fixes #330

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -2350,10 +2350,8 @@ HTTP/1.1 200 OK
 A client responds to this challenge by presenting the indicated URL for a human
 user to navigate to.  If the user chooses to complete this challenge (by visiting
 the website and completing its instructions), the client indicates this by
-sending a simple acknowledgement response to the server.
-
-type (required, string):
-: The string "oob-01"
+sending a simple acknowledgement response to the server.  The payload of this
+response is an empty JSON object ("{}", or "e30" base64url-encoded)
 
 ~~~~~~~~~~
 POST /acme/authz/1234/3
@@ -2367,9 +2365,7 @@ Content-Type: application/jose+json
     "nonce": "JHb54aT_KTXBWQOzGYkt9A",
     "url": "https://example.com/acme/authz/1234/3"
   }),
-  "payload": base64url({
-    "type": "oob-01"
-  }),
+  "payload": "e30",
   "signature": "Q1bURgJoEslbD1c5...3pYdSMLio57mQNN4"
 }
 ~~~~~~~~~~
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1179,7 +1179,7 @@ Content-Language: en
 ### External Account Binding
 
 The server MAY require a value to be present for the "external-account-binding"
-field.  This can be used to an ACME account with an existing account in a
+field.  This can be used to associate an ACME account with an existing account in a
 non-ACME system, such as a CA customer database.
 
 To enable ACME account binding, a CA needs to provision the ACME client with a
@@ -1517,7 +1517,7 @@ a virtual server starts up.
 
 In some cases, a CA running an ACME server might have a completely external,
 non-ACME process for authorizing a client to issue for an identifier.  In these
-case, the CA should provision its ACME server with authorization objects
+cases, the CA should provision its ACME server with authorization objects
 corresponding to these authorizations and reflect them as already valid in any
 orders submitted by the client.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1897,7 +1894,7 @@ Content-Language: en
 {
   "type": "urn:ietf:params:acme:error:unauthorized",
   "detail": "No authorization provided for name example.net",
-  "instance": "http://example.com/acme/doc/unauthorized"
+  "instance": "https://example.com/acme/doc/unauthorized"
```
cpu: I think this `instance` field should be removed instead of updated to use HTTPS ala https://github.com/ietf-wg-acme/acme/pull/355

draft-ietf-acme-acme.md
```diff
@@ -2002,7 +1999,13 @@ propagating across a cluster or firewall rules not being in place.
 
 Clients SHOULD NOT respond to challenges until they believe that the server's
 queries will succeed. If a server's initial validation query fails, the server
-SHOULD retry the query after some time.  While the server is still trying, the
+SHOULD retry the query after some time, in order to account for delay in setting 
+up responses such as DNS records or HTTP resources.  The precise retry schedule
+is up to the server, but server operators should keep in mind the operational
+scenarios that the schedule is trying to accommodate.  For example, a server
+might perform the first few of retries at intervals 5 or 10 seconds to catch
```
cpu: typo: "might perform the first few of retries". Maybe "might perform the first few retries" instead?

nit: "at intervals 5 or 10 seconds" -> "after an interval of 5 or 10 seconds" reads more naturally to me.

cpu: I also think the phrasing "to catch automated clients" is strange. Catch them doing what?

cpu: Also there's a double sapce before "For example, a server"

draft-ietf-acme-acme.md
```diff
@@ -2002,7 +1999,13 @@ propagating across a cluster or firewall rules not being in place.
...
+automated clients, then back off to intervals of a minute or more to deal with
+manual setup.  While the server is still trying, the
```
cpu: Overall I think this is a bit of a strange addition. Does the protocol really need to make explicit accommodations in a retry schedule for manual clients? the A in ACME makes me think otherwise. 

The feedback on list was "Any guidance on what "some times" means? At least in terms of a time scale, e.g., seconds, minutes, hours". I think it would be sufficient to give an example of 5 or 10 seconds and not elaborate much further.

draft-ietf-acme-acme.md
```diff
@@ -2347,10 +2350,8 @@ HTTP/1.1 200 OK
 A client responds to this challenge by presenting the indicated URL for a human
 user to navigate to.  If the user chooses to complete this challenge (by visiting
 the website and completing its instructions), the client indicates this by
-sending a simple acknowledgement response to the server.
-
-type (required, string):
-: The string "oob-01"
+sending a simple acknowledgement response to the server.  The payload of this
+response is an empty JSON object ("{}", or "e30" base64url-encoded)
```
cpu: Missing a `.` on the end of this sentence.

draft-ietf-acme-acme.md
bifurcation: SG, deleted.

draft-ietf-acme-acme.md
bifurcation: typo - fixed
"at intervals" - I think this is correct usage ("after an interval" would mean "wait 5-10s, then as quickly as you want")
"to catch" -> "to provide a prompt response for"

draft-ietf-acme-acme.md
bifurcation: Agreed.  Dialed back to say basically, "no more than every 5-10 seconds"

bifurcation:
Merging based on @cpu review

PR #359, opened 2017-11-28
Clarify external-account requirements

bifurcation:
While addressing Dale's Gen-ART comments, it occurred to me that while the document said that a CA could require external account binding, it didn't say *how*.  This PR adds a "meta" field and an error code to support that requirements proactively and reactively.


bifurcation:
Merging based on @cpu review

PR #360, opened 2017-11-30
Remove the Out-of-Band (OOB) Challenge type.

cpu:
The OOB challenge was added to accommodate non-ACME CAs building ACME
integrations. Since then two non-ACME CAs have implemented ACME
integrations using the external account binding feature but **not**
implementing OOB challenges. Since the target demographic has not
implemented this feature and has also not presented a use-case requiring
it there seems to be no strong reasons to keep it in-spec.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -2322,66 +2322,6 @@ If all of the above verifications succeed, then the validation is successful.
 If no DNS record is found, or DNS record and response payload do not pass these
 checks, then the validation fails.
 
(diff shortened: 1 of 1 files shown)


bifurcation:
@cpu If you could update this to remove the OOB challenge from the [IANA considerations](https://github.com/ietf-wg-acme/acme/blob/master/draft-ietf-acme-acme.md#validation-methods-iana-validation), I'll go ahead and merge

cpu:
@bifurcation Good catch - thanks! Addressed in 0753c73.

PR #362, opened 2017-12-01
Unifies to all object keys in camelCase

sophie-h:
Using camelCase like @bifurcation suggested. There does not seem to be standard among RFCs containing JSON. JWE uses snake_case, but I don't see a reason why the payload should use the same convention.

- Renamed sub-problems to subproblems since this spelling seems to be more common.
- Only changed the parts which refer to the keys. Phrases like "new-account request" are unchanged.
- Changes done via a script. I can resubmit the PR with a different convention if required.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -652,16 +652,16 @@ establish a new account with the server, prove control of an identifier, issue a
 certificate, and fetch an updated certificate some time after issuance.  The
 "->" is a mnemonic for a Location header pointing to a created resource.
 
-| Action               | Request                 | Response        |
-|:---------------------|:------------------------|:----------------|
-| Get a nonce          | HEAD newNonce           | 204             |
-| Create account       | POST newAccount         | 201 -> account  |
-| Submit an order      | POST newOrder           | 201 -> order    |
-| Fetch challenges     | GET  authz              | 200             |
-| Respond to challenge | POST challenge          | 200             |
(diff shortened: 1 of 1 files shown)


bifurcation:
@sophie-h Would you mind running your script again?  I think this is ready to merge, but needs some updates to account for the changes in #369 (sorry for adding conflicts!)

sophie-h:
Master is merged now. I noted one more small inconsistency: Order objects have authorizations, finalizeURL, and certificate as keys that refer to URL(s). Wouldn't it be more consistent to drop the URL from finalizeURL? Sorry for the pedantry.

cpu:
> Wouldn't it be more consistent to drop the URL from finalizeURL? Sorry for the pedantry.

I don't have strong opinions on that one way or the other :-)

bifurcation:
Big :+1: on "finalizeURL" -> "finalize"

@sophie-h I'll wait for you to make that change, then merge

sophie-h:
done

bmw:
I just wanted to chime in as a client developer and say that I prefer this approach to #363 as well. As pointed out above, this change is limited to the directory endpoint and the rest is in parts of the spec that I think most ACME software hasn't implemented yet.

In #363, the changes affect more pieces of existing functionality such as key authorizations and the new key field for accounts. I think this PR makes it easier for implementations to keep compatibility with new and old ACME software while everyone transitions and allows for more reuse of existing implementations.

PR #363, opened 2017-12-01
Unifies to all object keys in kebab-case

sophie-h:
kebab-case variant of #362


bifurcation:
Closing in preference to #362

PR #367, opened 2017-12-06
Add order finalization examples, fix new-order examples.

cpu:
Prior to this commit the new-order examples still included the "CSR"
field that was removed in favour of explicit order finalization.

This commit updates the new-order examples to include the "identifiers"
field instead & additionally adds an example finalization POST and an
example reply of a valid & completed order.

Thanks to @sophie-h for reporting this on-list initially.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1421,7 +1421,7 @@ Content-Type: application/jose+json
     "url": "https://example.com/acme/new-order"
   }),
   "payload": base64url({
-    "csr": "5jNudRx6Ye4HzKEqT5...FS6aKdZeGsysoCo4H9P",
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1448,7 +1448,6 @@ Location: https://example.com/acme/order/asdf
   "status": "pending",
   "expires": "2016-01-01T00:00:00Z",
 
-  "csr": "jcRf4uXra7FGYW5ZMewvV...rhlnznwy8YbpMGqwidEXfE",
```
sophie-h: There should be `identifiers` here, because this is an "object reflecting the client's request"

cpu: Good catch! You're right. I fixed this in 189d25e and also made sure the number of identifiers present matches the number of authorizations as well.

PR #369, opened 2017-12-11
Add a registry to track 'meta' fields

bifurcation:
Also, add some more "replace RFC XXXX with this document" notes, for clarity.

Thanks to @cpu for noticing that this was missing.


bifurcation:
Merging based on r+ from @cpu

PR #371, opened 2017-12-15
Small fixes in "Resources" section

sophie-h:


draft-ietf-acme-acme.md
```diff
@@ -657,16 +657,17 @@ establish a new account with the server, prove control of an identifier, issue a
...
-| Poll for status      | GET  authz          | 200             |
-| Check for new cert   | GET  cert           | 200             |
+| Action                | Request                   | Response       |
+|:----------------------|:--------------------------|:---------------|
+| Get directory         | GET  directory            | 200            |
+| Get nonce             | HEAD newNonce             | 204            |
+| Create account        | POST newAccount           | 201 -> account |
+| Submit order          | POST newOrder             | 201 -> order   |
+| Fetch challenges      | GET  order authorizations | 200            |
+| Respond to challenges | POST challenge urls       | 200            |
```
cpu: Boulder responds with a HTTPAccepted 202 to challenge POSTs - I don't see any guidance from the spec on this return code in my (quick) pass. 

I feel like both the challenge POST and the Finalize POST should probably return 202 instead of 200 but there will need to be text added to the draft in addition to this table.

@bifurcation Do you have any opinions about that?

bifurcation: I don't have any strong feelings here.  Redirect to @martinthomson for HTTP advice.

cpu: > Redirect to @martinthomson for HTTP advice.

301 or 302? :laughing:

martinthomson: Both are valid.  2xx would be a better choice here.  Throughout, aside from the 201 instances.  See http://httpwg.org/http-extensions/draft-ietf-httpbis-bcp56bis.html#http-status-codes

Note also that 204 for HEAD newNonce isn't necessary.  HEAD responses always don't have a body.

cpu: > Both are valid. 2xx would be a better choice here. Throughout, aside from the 201 instances. 

WFM. @sophie-h Can you update the table to use `2XX` where there isn't a 201?

> Note also that 204 for HEAD newNonce isn't necessary. HEAD responses always don't have a body.

Good call. [The portion of the spec](https://github.com/ietf-wg-acme/acme/blob/4053aa917c82d61add7a73837aa84260c0698d53/draft-ietf-acme-acme.md#getting-a-nonce) that mentions a 204 response for newNonce is specific to GETs to that endpoint. I think it makes the most sense to leave the newNonce action in the table a HEAD request and make the response `2xx` like the others rather than changing the action to GET and leaving the 204.

> p.s., that's a 303.

Noted :-)

sophie-h: @cpu counter proposal: fix the 204 consistently in a new PR an not let the sections diverge again

cpu: @sophie-h I'm not sure I follow - I think we want to allow GET requests to new-nonce, which does require a 204 response to indicate there shouldn't be a body in the GET response. I don't think anything has to change but the response code for the described HEAD request to new-nonce in this table, which shouldn't need a 204 per @martinthomson's comment.

Maybe I'm misunderstanding the proposed consistency fix you want to do in a separate PR?

sophie-h: @cpu 

```
7.2. Getting a Nonce
…
HEAD request to the new-nonce resource … SHOULD have status code 204 (No Content)
…
HEAD /acme/new-nonce HTTP/1.1
Host: example.com

HTTP/1.1 204 No Content
```

and maybe more

cpu: @sophie-h Aha! Thank you. I understand now. +1 on fixing those in a follow-up PR.

PR #373, opened 2017-12-19
Small changes to "Key Authorizations" section

sophie-h:
- I think without OOB all challenges use key-authorization
- "JWA" is not defined in this document. I think it can just be omitted

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -2053,7 +2053,7 @@ concatenating the token for the challenge with a key fingerprint, separated by a
 "." character:
 
 ~~~~~~~~~~
-  key-authorization = token || '.' || base64url(JWK_Thumbprint(accountKey))
+  keyAuthorization = token || '.' || base64url(JWK_Thumbprint(accountKey))
 ~~~~~~~~~~
 
 The "JWK\_Thumbprint" step indicates the computation specified in {{!RFC7638}},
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -2046,18 +2046,18 @@ need to specify which types of identifier they apply to.
...
-Several of the challenges in this document make use of a key authorization
+All challenges defined in this document make use of a key authorization
 string.  A key authorization is a string that expresses a domain holder's
 authorization for a specified key to satisfy a specified challenge, by
 concatenating the token for the challenge with a key fingerprint, separated by a
 "." character:
 
 ~~~~~~~~~~
-key-authz = token || '.' || base64url(JWK_Thumbprint(accountKey))
+  key-authorization = token || '.' || base64url(JWK_Thumbprint(accountKey))
```
shred: This should be `keyAuthorization`.

sophie-h: Thanks for the feedback @shred. I understood this line as an abstract definition of a term. Actually, it was never identical to the object key before. But it is the value for `keyAuthorization`. Maybe it would be helpful to use the same name?

shred: You are right, it is abstract and was never identical. But I think it would enhance readability, and there is also a camel-cased `accountKey` used in the same expression.

cpu: I agree with @shred - I think we should be consistent with `accountKey` and use `keyAuthorization` in the abstract definition.

shred:
@sophie-h Thank you for the change! :smile:

PR #375, opened 2017-12-27
[WIP] Changes to challenge response with empty objects

sophie-h:
- Respond with {} instead of providing keyAuthorization
- Remove "keyAuthorization" from the challenge
  object returned by the server. Fixes a potential misguide
  for clients that enabels a MitM attack.
- Keep the storing of the keyAuthorization on response
  to retain the same behavior if an account key roll-over
  happens before the challenge is validated

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -967,8 +967,7 @@ name validation.
       "type": "http-01",
       "status": "valid",
       "token": "DGyRejmCefe7v4NfDGDKfA"
-      "validated": "2014-12-01T12:05:00Z",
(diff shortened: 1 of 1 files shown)


jsha:
I like this change in theory, but it's substantive enough that I'd like to give the mailing list a couple more pokes to see if anyone has further thoughts on it before we land it.

richsalz:
This is not an editorial change, and I think it would require pushing back further on last-call timing.
Please bring this up and discuss on the mailing list; WG consensus will be required.

cpu:
> Please bring this up and discuss on the mailing list; WG consensus will be required.

This was brought up on the list Dec 27th and only positive responses were received.

PR #377, opened 2017-12-28
Make clear that contact is optional too

andriymahats:
Make clear contact field is optional as well. Isn't it correct?

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1007,7 +1007,7 @@ caching of this resource.
 
 A client creates a new account with the server by sending a POST request to the
 server's new-account URL.  The body of the request is a stub account object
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1007,8 +1007,7 @@ caching of this resource.
 
 A client creates a new account with the server by sending a POST request to the
 server's new-account URL.  The body of the request is a stub account object
-containing the "contact" field and optionally the "termsOfServiceAgreed"
-field.
+optionally containing "contact" and the "termsOfServiceAgreed" fields.
```
cpu: small nit: I think this reads clearer as "optionally containing **the** "contact" and "termsOfServiceAgreed" fields."

andriymahats: Thanks for correction, @cpu - fixed it

PR #383, opened 2018-01-04
Revocation example may use certificate key

joernheissler:
The example was misleading in case a certificate is revoked by signing the revocation request with the certificate's private key.

I assume "jwk" should be the public key from the certificate in this case and not the account key; see https://tools.ietf.org/html/rfc7515#section-4.1.3

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -340,19 +340,22 @@ JWS objects sent in ACME requests MUST meet the following additional criteria:
 * The JWS MUST NOT have a Message Authentication Code (MAC)-based algorithm in its "alg" field
 * The JWS Protected Header MUST include the following fields:
   * "alg" (Algorithm)
-  * "jwk" (JSON Web Key, only for requests to new-account and revoke-cert resources)
-  * "kid" (Key ID, for all other requests)
+  * "jwk" (JSON Web Key, for all requests not signed using an existing
+          account, e.g. newAccount)
+  * "kid" (Key ID, for all requests signed using an existing account)
   * "nonce" (defined in {{replay-protection}} below)
   * "url" (defined in {{request-url-integrity}} below)
 
 The "jwk" and "kid" fields are mutually exclusive. Servers MUST reject requests
 that contain both.
 
-For new-account requests, and for revoke-cert requests authenticated by certificate
-key, there MUST be a "jwk" field.
+For newAccount requests, and for revokeCert requests authenticated by certificate
+key, there MUST be a "jwk" field. This field MUST contain the public key
+corresponding to the private key used to sign the JWS.
 
-For all other requests, there MUST be a "kid" field. This field must
-contain the account URL received by POSTing to the new-account resource.
+For all other requests, the request is signed using an existing account and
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1926,7 +1926,7 @@ Content-Type: application/jose+json
 {
   "protected": base64url({
     "alg": "ES256",
-    "jwk": /* account key */,
+    "jwk": /* account or certificate key */,
```
cpu: I think this should just say `/* certificate key */` - I believe commenting the JWK as an "account key" was a mistake made before this PR. A revocation request authenticated by the account holder's private key should **not** include a JWK, but should include a KeyID. You can see how [Boulder splits up the KeyID case for account authorization and the JWK case for certificate keypair authorization](https://github.com/letsencrypt/boulder/blob/aa62a903036de8e6f79a991d0df4a964fd1abd05/wfe2/wfe.go#L771:L780) in the WFE2 code as one example implementation.

I think it might merit a second example that uses a "keyID" header since this is pretty unclear with just a JWK example :-)

@joernheissler Do you want to fix the comment in your PR and also add a keyID authenticated request with some explanatory accompanying text?

joernheissler: I agree that it would make more sense to use "kid" for account keys. But is it still possible to change the specs at this point? It might change meaning.

acme4j interpreted it the same way I did (Always send jwk):
https://github.com/shred/acme4j/blob/2ceffa42e38f45fd4c2e179508f5cee4e6753728/acme4j-client/src/main/java/org/shredzone/acme4j/Certificate.java#L210
https://github.com/shred/acme4j/blob/2ceffa42e38f45fd4c2e179508f5cee4e6753728/acme4j-client/src/main/java/org/shredzone/acme4j/connector/Connection.java#L84

https://ietf-wg-acme.github.io/acme/draft-ietf-acme-acme.html#rfc.section.6.2 contradicts itself though:
    
    “jwk” (JSON Web Key, only for requests to new-account and revoke-cert resources)
    “kid” (Key ID, for all other requests)

vs.

    For new-account requests, and for revoke-cert requests authenticated by
    certificate key, there MUST be a “jwk” field.

    For all other requests, there MUST be a “kid” field. This field must contain
    the account URL received by POSTing to the new-account resource.

cpu: > I agree that it would make more sense to use "kid" for account keys. But is it still possible to change the specs at this point? It might change meaning.

Yes it is still possible to change the spec. Changing this example as I described won't change meaning, it resolves an error.

> https://ietf-wg-acme.github.io/acme/draft-ietf-acme-acme.html#rfc.section.6.2 contradicts itself though:

I don't believe the two sections you shared contradict themselves. The second one says "and for revoke-cert requests **authenticated by certificate key**, there MUST be a “jwk” field." (emphasis mine). For requests authenticated by the account key a "kid" field is expected. 

I think there can be some clarification done but I don't think there is a contradiction.

bifurcation:
Merging because it looks like the issue raised by @cpu has been addressed.

PR #389, opened 2018-01-12
Add ALPN to TLS-SNI challenge

titanous:
This adds an ALPN requirement to the TLS-SNI challenge (now called tls-sni-03) in order to mitigate vulnerabilities in some shared hosting providers.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1760,7 +1760,7 @@ Link: <https://example.com/acme/some-directory>;rel="index"
       "token": "DGyRejmCefe7v4NfDGDKfA"
     },
     {
-      "type": "tls-sni-02",
+      "type": "tls-sni-03",
       "url": "https://example.com/acme/authz/1234/1",
       "token": "DGyRejmCefe7v4NfDGDKfA"
     },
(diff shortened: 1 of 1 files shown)


mhardeman:
It is altogether unclear how this mitigates any vulnerability in validations performed for domains on certain shared hosting providers.  Could you elaborate on that?

titanous:
@mhardeman Let's keep the discussion about this on the ACME mailing list, GitHub comments are only used for editorial changes.

billinghamj:
I'm not sure this is a great idea. We should build an ALPN-based approach from first principles, entirely unrelated to the existing specifics of the TLS-SNI challenge, rather than using it as a way to patch over the fundamental problems of the existing approach.

In my mind, if your process for proving hostname control involves setting the `Host` header or the SNI hostname to anything other than the hostname in question, it's really just asking for trouble. It's a fundamentally bad idea.

Why not make a new challenge where you initiate a valid TLS connection with the correct SNI hostname, but you use ALPN to indicate that you want to use an alternative protocol - `acme-validation` or whatever. That then enables the use of a custom purpose-built protocol where there can be no doubt about the intentions of the server, even if its ALPN implementation is broken (e.g. it repeats back `acme`).

This still provides all the benefits of the TLS-SNI approach, in that the ACME requests can be intercepted at the TLS-level, rather than the HTTP-level, but would do so in a way which is appropriate and explicit, rather than just scanning the current behaviour of the most popular websites, and hoping no more issues will come up.

cpu:
Given the state of the overall ACME draft w.r.t last call I think a substantial new challenge type should be a separate internet draft. @rolandshoemaker has [one such draft in-progress](https://datatracker.ietf.org/doc/draft-shoemaker-acme-tls-alpn/) (with a corresponding WG [mailing list thread](https://mailarchive.ietf.org/arch/msg/acme/o9QIj4Kv7QqwDe3nyJkl_w0lbwU/?qid=a7666acafdec3f8aa8a18cbcbe556151)). 

I'm going to close this PR and encourage folks to direct their attention to `draft-shoemaker-acme-tls-alpn`. Thanks all!

billinghamj:
@cpu So, to clarify, `tls-sni` challenges are being removed from the spec - leaving the `http` and `dns` ones only? Then any other additions will be considered for future specs

cpu:
@billinghamj That's correct. `TLS-SNI-01` was already removed from previously published drafts and master. `TLS-SNI-02` was removed with https://github.com/ietf-wg-acme/acme/commit/cfe118734bab3f6b7417463aad7e7faa8709df12 Only `DNS-01` and `HTTP-01` remain.

PR #390, opened 2018-01-12
Remove TLS-SNI-02 challenge type and associated refs.

cpu:
Recent developments[[0]] [[1]] [[2]] have identified real-world server/hosting
configurations that violate the assumptions of TLS-SNI-01 and its
currently specified replacement, TLS-SNI-02. In light of these issues
and the feasibility of addressing them across the entire Internet it
seems prudent that the ACME specification remove the affected
challenge type pending the development of a better alternative
(TLS-SNI-03?). This will allow the draft last-call to proceed while
the details of TLS-SNI-03 are worked out.

The "Default Virtual Hosts" sub-section of the "Operational
Considerations" section is removed since it spoke exclusively to
a concern with TLS-SNI-02 validation. When the dust has settled on
TLS-SNI-03 we should certainly include a section that describes why
TLS-SNI-01/02 were removed that could replace this information.

[0]: https://community.letsencrypt.org/t/2018-01-09-issue-with-tls-sni-01-and-shared-hosting-infrastructure/49996
[1]: https://community.letsencrypt.org/t/2018-01-11-update-regarding-acme-tls-sni-and-shared-hosting-infrastructure/50188
[2]: https://labs.detectify.com/2018/01/12/how-i-exploited-acme-tls-sni-01-issuing-lets-encrypt-ssl-certs-for-any-domain-using-shared-hosting/

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -514,7 +514,6 @@ in the "type" field (within the "urn:ietf:params:acme:error:" namespace):
 | caa                     | Certification Authority Authorization (CAA) records forbid the CA from issuing |
 | dns                     | There was a problem with a DNS query                                           |
 | connection              | The server could not connect to validation target                              |
-| tls                     | The server received a TLS error during validation                              |
 | incorrectResponse       | Response received didn't match the challenge's requirements                    |
 
 This list is not exhaustive. The server MAY return errors whose "type" field is
@@ -1760,11 +1759,6 @@ Link: <https://example.com/acme/some-directory>;rel="index"
       "token": "DGyRejmCefe7v4NfDGDKfA"
     },
     {
-      "type": "tls-sni-02",
-      "url": "https://example.com/acme/authz/1234/1",
-      "token": "DGyRejmCefe7v4NfDGDKfA"
-    },
-    {
       "type": "dns-01",
       "url": "https://example.com/acme/authz/1234/2",
       "token": "DGyRejmCefe7v4NfDGDKfA"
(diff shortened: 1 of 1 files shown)


cpu:
> Should there be any mention of tls-sni-02 that makes sure we use tls-sni-03 in a future RFC in case if gets re-added?

@kelunik I don't think we need a place-holder in-doc for this

draft-ietf-acme-acme.md
```diff
@@ -2709,7 +2603,6 @@ Initial Contents
 | Label      | Identifier Type | ACME | Reference |
 |:-----------|:----------------|:-----|:----------|
 | http-01    | dns             | Y    | RFC XXXX  |
-| tls-sni-02 | dns             | Y    | RFC XXXX  |
```
bifurcation: I agree with @kelunik that it would be good to have a tombstone here.  The standard practice is to have a "reserved" entry in the registry, so that you don't conflict later, but you also don't define it.  Here, I think the right row would be:

```
| tls-sni-02 | RESERVED | N | N/A |
```

cpu: @bifurcation Shouldn't there be one for `tls-sni-01` by the same logic? Should I add both in this PR?

cpu: I added both in 0bb71a0 - I can drop one if there's a reason we should only have a tombstone for tls-sni-02.

cpu: Confirmed OOB this addresses @bifurcation's comment.

PR #392, opened 2018-01-19
Clarify account URL == kid in account creation section.

cpu:
Section 6.2 "Request Authentication" says:
> For all other requests, there MUST be a "kid" field.  This field
> must contain the account URL received by POSTing to the new-account
> resource.

Prior to this commit Section 7.3 "Account Creation" doesn't mention that
the account URL also doubles as the key ID. To make the question of
"What is my Key ID?" more obvious for client implementers this commit
adds a brief mention of this fact where the specification discusses the
account URL.

Thanks to @unixcharles for flagging this and providing a chance to avoid
confusion and add clarity.


unixcharles:
thanks

PR #398, opened 2018-03-02
Add an error code to deal with content negotiation failure

bifurcation:
This is a follow-on to #395.  That PR was missing an error code to deal with content negotiation failure.


joernheissler:
See my comment on the list.

bifurcation:
Superseded by #410

PR #400, opened 2018-03-02
Status Clarifications

### LINKED ISSUE #391: authz status possible values incorrect
Hello,

I believe this is an editorial issue.

Under `7.1.4 Authorization Objects`:

> status (required, string): : The status of this authorization. Possible values are: "pending", "processing", "valid", "invalid" and "revoked".

I believe there are two inaccuracies:

* `deactivated` is not present in the list
* `processing` seems to be nonsensical for an authz and should not be present in the list

### LINKED ISSUE #397: Remove status on challenge objects
Right now, authorizations and challenges each have statuses, but they are mandatorily the same, since an authz becomes valid if and only if one of its challenges becomes valid. We should simplify by removing the copy of status on the challenge object.

We've seen some bugs in the wild related to separating these.

bifurcation:
This PR attempts to add a lot more clarity to the "status" field on objects and how it evolves.

* Adds a section describing the state transitions for the various object types. (This turns out to be a handy protocol summary as well)
* Aligns the valid values in the "status" field definitions to the description in that section
* Adds a "ready" state on orders to indicate when the order is ready to be finalized

Fixes #391
Fixes #397 (by showing why challenges need separate status)

Slightly beyond the scope of this PR, it might also be good to:
* Define an "expired" status for autz, independent of "invalid"
* Add a subsection of "Applying for Certificate Issuance" briefly noting / illustrating finalization

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -843,8 +843,8 @@ the server requires the client to complete, and any certificates that have
 resulted from this order.
 
 status (required, string):
-: The status of this order.  Possible values are: "pending", "processing",
-"valid", and "invalid".
+: The status of this order.  Possible values are: "pending",
+"ready", "processing", "valid", and "invalid".
 
 expires (optional, string):
 : The timestamp after which the server will consider this order invalid, encoded
@@ -951,8 +951,8 @@ identifier (required, object):
   : The identifier itself.
 
 status (required, string):
-: The status of this authorization.  Possible values are: "pending", "processing",
-"valid", "invalid" and "revoked".
+: The status of this authorization.  Possible values are: "pending",
+"valid", "invalid", "deactivated", and "revoked".
 
 expires (optional, string):
 : The timestamp after which the server will consider this authorization invalid,
@@ -1036,7 +1036,7 @@ the challenges listed in the authorization transitions to the
 state.  If there is an error while the authorization is still
 pending (e.g., the authorization times out), then the authorizatoin
 transitions to the "invalid" state.  Once the authorization is in
-the valid state, it can expire ("expired"), be deactivated by the
+the valid state, it can expire ("invalid"), be deactivated by the
 client ("deactivated", see {{deactivating-an-authorization}}),
 revoked by the server ("revoked").
 
@@ -1047,16 +1047,18 @@ Error        |  Challenge -> valid
    +---------+---------+
    |                   |
    V                   V
-invalid              valid
-                       |
+invalid <----------- valid
+            Time       |
+           after       |
+         "expires"     |
                        |
            +-----------+-----------+
-           |           |           |
-     Time  |           |           |
-    after  |    Client |    Server |
- "expires" |   deactiv.|    revoke |
-           V           V           V
-        expired   deactivated   revoked
+           |                       |
+           |                       |
+    Client |                Server |
+   deactiv.|                revoke |
+           V                       V
+      deactivated               revoked
 ~~~~~~~~~~
 
 Order objects are created in the "pending" state.  Once all of the
@@ -2165,8 +2167,8 @@ url (required, string):
 : The URL to which a response can be posted.
 
 status (required, string):
-: The status of this challenge.  Possible values are: "pending", "valid",
-and "invalid".
+: The status of this challenge.  Possible values are: "pending",
+"processing", "valid", and "invalid".
 
 validated (optional, string):
 : The time at which the server validated this challenge, encoded in the
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1000,6 +1000,111 @@ name validation.
 }
 ~~~~~~~~~~
 
+### Status Changes
+
+Each ACME object type goes through a simple state machine over its
+lifetime.  The "status" field of the object indicates which state
+the object is currently in.
+
+Challenge objects are created in the "pending" state.  They
+transition to the "processing" state when the client responds to the
+challenge (see {{responding-to-challenges}}) and the server begins
+attempting to validate that the client has completed the challenge.
+If validation is successful, the challenge moves to the "valid"
+state; if there is an error, the challenge moves to the "invalid"
+state.
+
+~~~~~~~~~~
+         pending
+            | 
+            | Receive
+            | response
+            V
+        processing
+            |
+Successful  |   Failed
+validation  |   validation
+  +---------+---------+
+  |                   |
+  V                   V
+valid              invalid
+~~~~~~~~~~
+
+Authorization objects are created in the "pending" state.  If one of
+the challenges listed in the authorization transitions to the
+"valid" state, then the authorization also changes to the "valid"
+state.  If there is an error while the authorization is still
+pending (e.g., the authorization times out), then the authorizatoin
+transitions to the "invalid" state.  Once the authorization is in
+the valid state, it can expire ("expired"), be deactivated by the
+client ("deactivated", see {{deactivating-an-authorization}}),
+revoked by the server ("revoked").
+
+~~~~~~~~~~
+          pending
+             |
+Error        |  Challenge -> valid
+   +---------+---------+
+   |                   |
+   V                   V
+invalid              valid
+                       |
+                       |
+           +-----------+-----------+
+           |           |           |
+     Time  |           |           |
+    after  |    Client |    Server |
+ "expires" |   deactiv.|    revoke |
+           V           V           V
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1000,6 +1000,120 @@ name validation.
...
+
+Challenge objects are created in the "pending" state.  They
+transition to the "processing" state when the client responds to the
+challenge (see {{responding-to-challenges}}) and the server begins
+attempting to validate that the client has completed the challenge.
+Note that within the "processing" state, the server may attempt to
+validate the challenge multiple times (see {{retrying-challenges}}).
+Likewise, client requests for retries do not cause a state change.
+If validation is successful, the challenge moves to the "valid"
+state; if there is an error, the challenge moves to the "invalid"
```
cpu: If there is an error, and the challenge moves to the "invalid" stat, how does the client request retry? That seems like the primary reason you would want to retry and there isn't any information about how the server should handle failures during processing except to change out of processing.

draft-ietf-acme-acme.md
```diff
@@ -1000,6 +1000,120 @@ name validation.
...
+  |                   |
+  V                   V
+valid              invalid
+~~~~~~~~~~
+
+Authorization objects are created in the "pending" state.  If one of
+the challenges listed in the authorization transitions to the
+"valid" state, then the authorization also changes to the "valid"
+state.  If there is an error while the authorization is still
+pending (e.g., the authorization times out), then the authorizatoin
```
cpu: what does "the authorization times out" mean? I thought perhaps expiry but you mention that distinctly later in this paragraph.

draft-ietf-acme-acme.md
```diff
@@ -1000,6 +1000,120 @@ name validation.
...
+transitions to the "invalid" state.  Once the authorization is in
+the valid state, it can expire ("invalid"), be deactivated by the
+client ("deactivated", see {{deactivating-an-authorization}}),
```
cpu: I think you're missing a connecting word here between deactivating and revoking.

> or, revoked by the server ("revoked").

draft-ietf-acme-acme.md
```diff
@@ -1000,6 +1000,120 @@ name validation.
...
+invalid <----------- valid
+            Time       |
+           after       |
+         "expires"     |
+                       |
+           +-----------+-----------+
+           |                       |
+           |                       |
+    Client |                Server |
+   deactiv.|                revoke |
```
cpu: nit: If you put the revocation case on the left fork I think you could fit the whole word "deactivated" on the right fork

draft-ietf-acme-acme.md
```diff
@@ -1000,6 +1000,120 @@ name validation.
...
+      deactivated               revoked
+~~~~~~~~~~
+
+Order objects are created in the "pending" state.  Once all of the
+authorizations listed in the order object are in the "valid" state,
+the order transitions to the "ready" state.  The order moves to the
+"processing" state after the client submits a request to the order's
+"finalize" URL and the CA begins the issuance process for the
+certificate.  Once the certificate is issued, the order enters the
+"valid" state.  If a fatal error occurs at any of these stages, the
```
cpu: I think we should skip the word "fatal" here - we never define what a fatal error is versus a normal error. They're all just errors.

draft-ietf-acme-acme.md
```diff
@@ -1000,6 +1000,120 @@ name validation.
...
+order moves to the "invalid" state.
```
cpu: This paragraph (& the flow chart) are also missing three ways that an order can become invalid:
1) One of its authorizations becomes invalid because it was deactivated
2) One of its authorizations becomes invalid because it expired
3) The order expires

draft-ietf-acme-acme.md
jsha: *authorizatoin->authorization

draft-ietf-acme-acme.md
```diff
@@ -1000,6 +1000,120 @@ name validation.
...
+revoked by the server ("revoked").
+
+~~~~~~~~~~
+          pending
+             |
+Error        |  Challenge -> valid
+   +---------+---------+
+   |                   |
+   V                   V
+invalid <----------- valid
```
jsha: Let's treat expired as its own state, and note that it doesn't show up in the "status" field: the URL just starts to 404.

Also, there should be an arrow from pending to expired.

draft-ietf-acme-acme.md
```diff
@@ -1000,6 +1000,120 @@ name validation.
...
+           |                       |
+    Client |                Server |
+   deactiv.|                revoke |
+           V                       V
+      deactivated               revoked
+~~~~~~~~~~
+
+Order objects are created in the "pending" state.  Once all of the
+authorizations listed in the order object are in the "valid" state,
+the order transitions to the "ready" state.  The order moves to the
```
jsha: The "ready" state makes some sense but is a significant addition. I'm inclined to leave it out of this change, to more clearly separate "additional documentation" from "protocol changes".

draft-ietf-acme-acme.md
jsha: > One of its authorizations becomes invalid because it was deactivated

Or the authorization was revoked. :-)

We should probably say "If any of the authorizations changes to a state other than pending or valid, the order becomes invalid."

draft-ietf-acme-acme.md
joernheissler: I think "invalid" is a final decision. If the client wants another try, they'd need to create a new order.
Section 8.2 says that "the server SHOULD retry the query after some time".
Being unable to retry individual challenges might mean that some users will have a hard time acquiring a certificate. But then they should just fix their setup and respond to the challenge only when it will work.
I.e. I agree with the above change.

draft-ietf-acme-acme.md
bifurcation: We don't require 404 when an authz expires:

>  If the client fails to complete the required actions before the "expires" time, then the server SHOULD change the status of the order to "invalid" and MAY delete the order resource.

So we shouldn't add any text here that would imply that 404 is necessary.

draft-ietf-acme-acme.md
bifurcation: Funny you should say that, because "expires" is also a new state :)  This change is not large in spec terms -- this PR is all it needs.  It doesn't seem like it should be a huge change for implementations either.

draft-ietf-acme-acme.md
bifurcation: Yeah, what @joernheissler said.  I had also been thinking of "invalid" as a final state.  ISTM if you want to signal to the client that something has gone wrong and a retry is needed, there are basically three options:

1. Keep it in "processing" and add to the subproblems array
2. Move it to "invalid" and add to the subproblems array
3. Move it to a new "retry" state and add to the subproblems array

(1) is what we do now, and in case you can't tell by my phrasing here, I don't think we need any more signal than adding to the subproblems array.  It seems like the only real argument for having a non-"processing" state is if find it distasteful to have "processing" with "error" populated, in which case I would propose we do (3) instead.

cpu: Ok. That sounds reasonable.

draft-ietf-acme-acme.md
```diff
@@ -1705,12 +1705,16 @@ action the client should take:
   requirements.  Check the "authorizations" array for entries that are still
   pending.
 
-* "processing": The server agrees that the requirements have been fulfilled, and
-  is in the process of generating the certificate.  Retry after the time given
-  in the "Retry-After" header field of the response, if any.
+* "ready": The server agrees that the requirements have been
+  fulfilled, and is awaiting on finalization.  Submit a finalization
```
cpu: nit: "and is awaiting on finalization" -> "and is awaiting finalization".

draft-ietf-acme-acme.md
```diff
@@ -1705,12 +1705,16 @@ action the client should take:
...
+  request.
+
+* "processing": The certificate is being issued.  Retry after the
```
cpu: "Retry" in this case might need to be disambiguated slightly, what about "Poll the order again after the ... " ?

jsha:
I'm fine accepting the "ready" state based on the above, although Let's Encrypt won't have it implemented by v2 launch time. I think that's fine. One tiny bit of further clarification: I was thinking of "expired" as a meta-state that is not necessarily reflected in the status field. I think we can just add a sentence saying something like "objects in an expired state may simply return 404."

PR #401, opened 2018-03-02
Clarify that pre-authorization allows wildcard identifiers.

cpu:
The `newOrder` endpoint has specific text that describes that a new
order request may carry DNS type identifiers with a wildcard prefix.

There's no reason that a preauthorization request DNS type identifier
can't have a wildcard prefix so this commit duplicates that text here as
well.

Originally flagged by @bmw, thanks!

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1641,6 +1641,14 @@ To request authorization for an identifier, the client sends a POST request to
 the new-authorization resource specifying the identifier for which authorization
 is being requested.
 
+Any identifier of type "dns" in a pre-authorization request MAY have a wildcard
+domain name as its value. A wildcard domain name consists of a single asterisk
+character followed by a single full stop character ("\*.") followed by a domain
+name as defined for use in the Subject Alternate Name Extension by RFC 5280
+{{!RFC5280}}. An authorization returned by the server for a wildcard domain name
+identifier MUST NOT include the asterisk and full stop ("\*.") prefix in the
+authorization identifier value.
+
 identifier (required, object):
 : The identifier that the account is authorized to represent:
 


cpu:
I think this is strictly editorial - there was no language forbidding this for pre-authorization, it just wasn't explicitly permitted like it probably should have been. I don't plan to raise this on-list.

bifurcation:
So the flow here would be as follows?

* Client sends request for authz for "*.example.com"
* Server responds with authz for "example.com" + DNS challenge

I don't love it.  I think I would rather require that preAuth require that the identifiers in the request and response MUST be the same, and require that wildcard issuance go through newOrder.

cpu:
> Server responds with authz for "example.com" + DNS challenge

I think that's 100% a policy decision. The spec doesn't say anything about validating wildcard identifiers with DNS challenges.

> I don't love it. I think I would rather require that preAuth require that the identifiers in the request and response MUST be the same, and require that wildcard issuance go through newOrder.

I'm 1000% fine with this too. I leave it to people that actually want to use pre-authorization to defend that it should support wildcard identifiers.

On the other hand, as someone that doesn't want pre-authorization to be a thing: Why have a less powerful version of newOrder laying around? :-P

bifurcation:
I talked with @ekr about this, as someone who has advocated for preauthorization before, and he was pretty meh on this.  So I would propose we close it.

cpu:
@bifurcation I don't think we should close this leaving the text ambiguous about whether pre-authorization supports wildcard identifiers. I would support adding text that says the pre-authorization identifiers MUST be the same as the authorizations and only newOrder supports wildcard OR adding the text in this PR.

bifurcation:
Replaced by #412

PR #402, opened 2018-03-02
Add optional "wildcard" field to authorizations.

cpu:
`NewOrder` and `NewAuthz` pre-authorization requests are allowed
to send DNS type identifiers with a value that contains a wildcard
prefix ('*') as the leftmost label.

The authorization object created for the wildcard identifiers in the
requests are **not** allowed to carry the wildcard prefix in the
authorization identifier per spec language:
> An authorization returned by the server for a wildcard domain name 
> identifier MUST NOT include the asterisk and full stop ("\*.") prefix in the
> authorization identifier value

This is sensible because the authorization is not for the logical 
wildcard but for the specific base domain.

This leaves an ambiguous situation for client developers/users. Imagine
a situation where a user requests a new order with the following
request body:
```
{
  "identifiers": [
    { "type": "dns", "value": "example.com" },
    { "type": "dns", "value": "*.example.com" }
  ]
}
```
It's possible two authorizations will be created for this case, but both
will contain the identifier `{ "type":"dns", "value": "example.com"}`.
This can be confusing to users! (e.g.  shred/acme4j#53)

If one of the authorizations fails a challenge validation operation the
client can't know whether it was the "example.com" validation, or the
"*.example.com" validation that failed and won't be able to prompt the
user to retry with the failed domain removed. Generally speaking it
makes presenting accurate error information difficult without
over-fitting Let's Encrypt's particular Wildcard issuance policies.
(E.g. see discussion in https://github.com/certbot/certbot/pull/5620#issuecomment-369111947 and 
https://github.com/certbot/certbot/issues/5613#issuecomment-368097551)

This commit introduces an optional boolean authorization field that can
be used to indicate the authorization's identifier should be considered
the base domain of a wildcard identifier that was requested.

**Note to reviewers: This builds on top of #401 and only 050474a is new
content.**

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -917,7 +917,8 @@ character followed by a single full stop character ("\*.") followed by a domain
 name as defined for use in the Subject Alternate Name Extension by RFC 5280
 {{!RFC5280}}. An authorization returned by the server for a wildcard domain name
 identifier MUST NOT include the asterisk and full stop ("\*.") prefix in the
-authorization identifier value.
+authorization identifier value. The returned authorization MUST have the
+optional "wildcard" field set to true.
 
 The elements of the "authorizations" and "identifiers" array are immutable once
 set.  The server MUST NOT change the contents of either array after they are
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1641,6 +1654,15 @@ To request authorization for an identifier, the client sends a POST request to
 the new-authorization resource specifying the identifier for which authorization
 is being requested.
 
+Any identifier of type "dns" in a pre-authorization request MAY have a wildcard
```
bifurcation: Let's please change this `MAY` to `MUST NOT` and clean up the text that follows.  Preauth is for authorizing specific identifiers; each authz corresponds to a single identifier, so the identifier in the request should equal the one in the authz.  With wildcards, there's ambiguity about what identifier will be in the authz, or whether multiple authzs might be needed.  Let's just avoid that complexity.

cpu: > Let's please change this MAY to MUST NOT

Sure, sounds good.

>  Preauth is for authorizing specific identifiers;

I don't think this is clear from the text that exists today. I'll remove everything but the "Wildcard" authz field text from this PR since this is feedback for https://github.com/ietf-wg-acme/acme/pull/401 but I [stand by my comment there](https://github.com/ietf-wg-acme/acme/pull/401#issuecomment-370148270): You and EKR have a notion of what pre-authorization can be used for that isn't reflected in the text. There should be a PR and a mailing list discussion about introducing text that clarifies the pre-authorization restrictions compared to newOrder.

cpu:
Closing this to resubmit without the baggage of #401.

PR #405, opened 2018-03-04
Editorial changes

joernheissler:


bifurcation:
Note: There is one "should" -> "SHOULD" here, but I don't think that's substantive enough to hold up the PR.

PR #407, opened 2018-03-04
Explicitly forbid to answer GET requests for account info.

joernheissler:
Section "Request Authentication" already forbade it.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1189,7 +1189,7 @@ Content-Type: application/jose+json
 
 ### Account Information
 
-Servers SHOULD NOT respond to GET requests for account resources as these
+Servers MUST NOT respond to GET requests for account resources as these
 requests are not authenticated.  If a client wishes to query the server for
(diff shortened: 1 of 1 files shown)


bifurcation:
In general, we have not been making MUST-level requirements for servers in this document because CAs are subject to a lot of external constraints already.  Under that theory, I would say we should just close this.  @cpu @jsha wdyt?

joernheissler:
If you don't want this change, can you change "Servers MUST NOT respond to GET requests for resources that might be considered sensitive. Account resources are the only sensitive resources defined in this specification." instead to resolve the inconsistency?

cpu:
@bifurcation I agree with @joernheissler - this isn't a new MUST-level requirement, it's already said elsewhere that the server MUST NOT respond to sensitive resources and the account object has contact information you wouldn't want returned to an unauthenticated GET request. It seems fine to emphasize this in the account information section.

cpu:
I'm going to merge this. @bifurcation gives an OOB +1.

Thanks @joernheissler !

PR #409, opened 2018-03-05
Reword to address ambiguity pointed out by @joernheissler

### LINKED ISSUE #404: Wording in Introduction
> For example, the external
account binding feature (see {{external-account-binding}}) can be used to
associate authorizations with an account that were not validated through the
ACME authorization process.

To me (non-native English speaker) "were not" refers to "an account" which would be a grammar mistake. If it refers to "authorizations" would it be better to switch "authorizations" and "an account"?

bifurcation:
Fixes #404


bifurcation:
Merging because editorial.

PR #403, opened 2018-03-03
Clarify serialization requirements for root and nested JWS messages

uhhhh2:
# What is Included
- An error code to deal with JWS serialization failures (copied from https://github.com/ietf-wg-acme/acme/pull/398) 
- A definition of a "root JWS" (a JWS encapsulating an entire ACME request payload) 
- A definition of a "nested JWS" (a JWS included within an ACME request payload)
- A modified "application/jose" JWS serialization Content-Type (to include specifications for both root and nested JWS instances)
- A modified "application/jose+json" JWS serialization Content-Type (to include specifications for both root and nested JWS instances)
- A "nested JWS" field type option in the existing ACME registries (to denote a nested JWS whose actual JSON type depends on the requirements of the JWS serialization Content-Type)
- A new ACME registry for the JWS serialization Content-Types

# What is NOT Included
- A guide for converting between JWS serialization formats before and/or after using a pre-existing JWS implementation (see https://github.com/uhhhh2/jwe-jws-serialization-conversion-guide)

PR #410, opened 2018-03-05
Require use of the Flattened JSON Serialization

bifurcation:
This is my attempt to find a quick way out of this content-type mess.  Replaces #403 and #398


jsha:
This also looks good to me, thanks for drafting it!

PR #414, opened 2018-03-13
Typo fix in examples

nzatnsr:
Change "type:"dns" to "type":"dns"


draft-ietf-acme-acme.md
```diff
@@ -1602,7 +1602,10 @@ Content-Type: application/jose+json
     "url": "https://example.com/acme/new-order"
   }),
   "payload": base64url({
-    "identifiers": [{"type:"dns","value":"example.com"}],
+    "identifiers": [{"type": "dns","value": "example.com"}],
```
cpu: I think you need to delete line 1605 since you duplicated it in a more consistent manner in lines 1606..1608

PR #415, opened 2018-03-20
Clarify questions about account objects.

cpu:
1) Include optional `externalAccountBinding` field in the Account
   Creation section with a forward pointer to the external account
   binding section.
2) Include the mandatory "orders" field in the example new account
   response.
3) Correct the "orders" field type in Section 9.7.1 to match the type
   from 7.1.2

Thanks to @nzatnsr for flagging these issues on the WG Mailing List.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1265,7 +1265,7 @@ Link: <https://example.com/acme/some-directory>;rel="index"
     "mailto:admin@example.com"
   ],
 
-  "orders": "https://example.com/acme/orders/acct/1"
+  "orders": "https://example.com/acme/acct/1/orders"
 }
 ~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1259,7 +1263,9 @@ Link: <https://example.com/acme/some-directory>;rel="index"
   "contact": [
     "mailto:cert-admin@example.com",
     "mailto:admin@example.com"
-  ]
+  ],
+
+  "orders": "https://example.com/acme/orders/acct/1"
```
rolandshoemaker: In the `Account objects` section we use a slightly different form for the `orders` URL which probably makes slightly more sense: `"orders": "https://example.com/acme/acct/1/orders"` (i.e. list of orders for acct `1`). Not that important but they should probably be synced.

cpu: Good catch, I agree that this format makes more sense and being consistent is the best path forward. Fixed in 17f466a

PR #420, opened 2018-04-11
Comments from Tim Hollebeek

bifurcation:


draft-ietf-acme-acme.md
```diff
@@ -317,9 +317,10 @@ their TLS implementations.  ACME servers that support TLS 1.3 MAY allow clients
 to send early data (0-RTT).  This is safe because the ACME protocol itself
 includes anti-replay protections (see {{replay-protection}}).
 
-ACME clients SHOULD send a User-Agent header in accordance with
-{{!RFC7231}}, including the name and version of the ACME software in
-addition to the name and version of the underlying HTTP client software.
+ACME clients MUST send a User-Agent header, in accordance with
+{{!RFC7231}}. this header SHOULD include the name and version of the
```
martinthomson: Capital after period.

draft-ietf-acme-acme.md
```diff
@@ -3052,6 +3053,27 @@ There are certain factors that arise in operational reality that operators of
...
 
+## Key Selection
+
+ACME relies on two different classes of key pair:
+
+* Account key pairs, which are used to authenticate account holders
+* Certificate key pairs, which are used to sign and verify CSRs (and whose
+  public keys are included in certificates)
+
+Of these two types of key pair, account key pairs are far more dangerous.  While
```
martinthomson: Saying a key is dangerous is weird.  "Compromise an account key has more serious consequences than compromise of a certificate key." might be a better first sentence.

draft-ietf-acme-acme.md
```diff
@@ -3052,6 +3053,27 @@ There are certain factors that arise in operational reality that operators of
...
+the compromise of a certificate key pair allows the attacker to impersonate the
+entities named in the certificate for the lifetime of the certificate, the
+compromise of an account key pair allows the attacker to take full control of
+the victim's ACME account -- potentially causing unintended issuance or
+revocation.
```
martinthomson: It might be best to mention what capabilities are of concern:
1. the ability to revoke existing certificates
2. the ability to issue certificates using existing authorizations

Are these the main ones?

draft-ietf-acme-acme.md
```diff
@@ -3052,6 +3053,27 @@ There are certain factors that arise in operational reality that operators of
...
+
+For this reason, it is RECOMMENDED that account key pairs be used for no other
+purpose besides ACME authentication.  For example, the public key of an account
+key pair SHOULD NOT be included in a certificate.  ACME clients and servers
+SHOULD verify that a CSR submitted in a finalize request does not contain a
+public key for any known account key pair.
```
martinthomson: I get that the global lookup on keys might be costly, so you don't want to MUST this, but is there any reason not to put a MUST on the account key associated with the finalize request?

bifurcation: Just for background: Typically, we have avoided placing MUSTs on servers in places where a server's local policy might reasonably override the MUST.  For example, for a long time, we didn't specify which keys a server MUST trust for revocation.  

In the revocation case, I think people agreed to the MUST because servers could just make their own custom revocation thing if they disagreed with the MUST.  Here, that doesn't really apply, but I think there's a real security requirement here.

bifurcation:
Thanks for the comments @martinthomson.  I've updated the PR to address.

cpu:
There has been list discussion on this PR and consensus is positive. I'm going to merge this now that @bifurcation has addressed the feedback provided.

Issue #419, opened 2018-04-10
Order of identifiers and authorizations in new-order is not specified


csware:
On the `new-order` reuqest, there are two arrays returned, one for the `identifiers` and one for the `authorizations`. However, there is no order specified in ACMEv2. As two arrays are returned (and no hash), this might imply that the order of `identifiers` matches the order of `authorizations`.

This is not specified right now. Having this specified would allow clients to know in advance for which identity they are requesting the authorization challenge before requesting the URL, e.g. for better error reporting.

Also ok would be to just explicitly specify that the order is not guaranteed or the number of entries is not guaranteed to match in order to make it more clear to developers.

cf. https://community.letsencrypt.org/t/dns-based-validation-fails-on-renew/59027?u=mrtux

jsha:
From the outcome of the linked conversation, we should just specify that no specific order is guaranteed.

cpu:
> we should just specify that no specific order is guaranteed.

:+1: - https://github.com/ietf-wg-acme/acme/pull/421

I will start a mailing list thread since this is a new SHOULD NOT.

PR #421, opened 2018-04-17
Clients SHOULD NOT assume ident/authz/chal sort order.

cpu:
This PR clarifies that a client SHOULD NOT assume that the "identifiers", "authorizations" or "challenges" arrays returned in various server response objects are sorted.

This PR also clarifies that while some [popular ACME server implementations](https://github.com/letsencrypt/boulder) have a 1:1 relationship between order identifiers and order authorizations this isn't a universal truth.

This resolves https://github.com/ietf-wg-acme/acme/issues/419

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -869,7 +869,7 @@ This field is structured as a problem document {{!RFC7807}}.
 authorizations (required, array of string):
 : For pending orders, the authorizations that the client needs to complete
 before the requested certificate can be issued (see
-{{identifier-authorization}}).  For final orders (in the "valid" or "invalid" state), the authorizations that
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -869,7 +869,7 @@ This field is structured as a problem document {{!RFC7807}}.
 authorizations (required, array of string):
 : For pending orders, the authorizations that the client needs to complete
 before the requested certificate can be issued (see
-{{identifier-authorization}}).  For final orders (in the "valid" or "invalid" state), the authorizations that
+{{identifier-authorization}}). The authorizations required are dictated by server policy and there may not be a 1:1 relationshop between the order identifiers and the authorizations required. For final orders (in the "valid" or "invalid" state), the authorizations that
```
jsha: *relationship

cpu: Good catch, thanks! Fixed in 3560d16

cpu:
@jsha Is 6 days without any feedback on-list sufficient to merge?

PR #424, opened 2018-04-24
Dale Worley's Gen-ART review

bifurcation:
A few editorial things that Dale found.


cpu:
Editorial diffs only. Merging as-is. Thanks @bifurcation !

PR #425, opened 2018-05-15
Responses to EKR's AD review comments

bifurcation:


draft-ietf-acme-acme.md
```diff
@@ -1482,35 +1482,38 @@ order to recover from a key compromise or proactively mitigate the impact of an
...
 
-newKey (required, JWK):
-: The JWK representation of the new key
+oldKey (required, JWK):
+: The JWK representation of the old key
 
 The client then encapsulates the key-change object in an "inner" JWS, signed with the
-requested new account key (i.e., the key matching the "newKey" value).  This JWS
-then becomes the payload for the "outer" JWS that is the body of the ACME
+requested new account key (i.e., the key matching the "newKey" value).
```
sophie-h: Usage of obsolete `newKey` field

draft-ietf-acme-acme.md
```diff
@@ -154,6 +155,9 @@ deploying an HTTPS server using ACME, the experience would be something like thi
   issue a certificate for the intended domain name(s).
 * The CA verifies that the client controls the requested domain name(s) by
   having the ACME client perform some action related to the domain name(s).
+  For example, the CA might require a client requsting example.com
```
pcworld: requ**e**sting

draft-ietf-acme-acme.md
```diff
@@ -543,7 +543,6 @@ in the "type" field (within the "urn:ietf:params:acme:error:" namespace):
 | rateLimited             | The request exceeds a rate limit                                               |
 | rejectedIdentifier      | The server will not issue for the identifier                                   |
 | serverInternal          | The server experienced an internal error                                       |
-| tls                     | The server received a TLS error during validation                              |
```
ilaril: TLS errors can still happen if HTTP-01 redirects to https:// and TCP connection to port 443 succeeds, but the TLS handshake fails.

draft-ietf-acme-acme.md
```diff
@@ -871,7 +905,7 @@ This field is structured as a problem document {{!RFC7807}}.
 authorizations (required, array of string):
 : For pending orders, the authorizations that the client needs to complete
 before the requested certificate can be issued (see
-{{identifier-authorization}}). The authorizations required are dictated by server policy and there may not be a 1:1 relationship between the order identifiers and the authorizations required. For final orders (in the "valid" or "invalid" state), the authorizations that
+{{identifier-authorization}}), including authorizations that the client has completed in the past. The authorizations required are dictated by server policy and there may not be a 1:1 relationship between the order identifiers and the authorizations required. For final orders (in the "valid" or "invalid" state), the authorizations that
```
cpu: Maybe slightly clearer as "Including unexpired authorizations that the client has completed in the past for identifiers specified in the order"?

draft-ietf-acme-acme.md
```diff
@@ -1273,9 +1324,9 @@ Link: <https://example.com/acme/some-directory>;rel="index"
 
 ### Finding an Account URL Given a Key
 
-If the server already has an account registered with the provided account key,
+If the server receives a newAccount request signed with a key for which it already has an account registered with the provided account key,
 then it MUST return a response with a 200 (OK) status code and provide the URL of
-that account in the Location header field.  This allows a client that has
+that account in the Location header field.  The body of this response is the account value as it existed on the server before this request; any fields in the request object MUST be ignored.  This allows a client that has
```
cpu: What is an "account value"? The ID/URL? I think the spec uses "Account object" to mean the literal account resource elsewhere no?

draft-ietf-acme-acme.md
```diff
@@ -1699,13 +1763,25 @@ Content-Type: application/jose+json
 
 The CSR encodes the client's requests with regard to the content of the
 certificate to be issued.  The CSR MUST indicate the exact same set of requested
-identifiers as the initial new-order request, either in the commonName portion
+identifiers as the initial new-order request.  Identifiers of type "dns" MUST apper in either in the commonName portion
```
cpu: typo: "MUST apper" -> "appear"

typo 2: "MUST apper in either in the" -> "MUST appear in either the"

draft-ietf-acme-acme.md
```diff
@@ -1699,13 +1763,25 @@ Content-Type: application/jose+json
...
 of the requested subject name, or in an extensionRequest attribute {{!RFC2985}}
-requesting a subjectAltName extension.
+requesting a subjectAltName extension.  (These identifiers may appear
+in any order.)  Specifications that define
```
cpu: fwiw: I tried to always say "sort order" when talking about this to avoid conflating with ACME Orders. I think it might help clarify to do the same here.

draft-ietf-acme-acme.md
```diff
@@ -1699,13 +1763,25 @@ Content-Type: application/jose+json
...
+new identifier types must specify where in the certificate these
```
cpu: I think where it says "where in the certificate these identifiers can appear" should be "where in the certificate signing request these identifiers can appear" no?

draft-ietf-acme-acme.md
```diff
@@ -2804,6 +2880,10 @@ interoperable definition and does not overlap with existing validation methods.
 That is, it should not be possible for a client and server to follow the
 same set of actions to fulfill two different validation methods.
 
+The values "tls-sni-01" and "tls-sni-02" are reserved because they
+were used in pre-RFC versions of this specification to denote
+validation methods that were found not to be secure in some cases.
```
cpu: I think it would be worth adding "and were removed from the specification" to the end of this paragraph.

cpu:
>  Can you explain why the newKey/oldKey changes were required? I think that should be pulled out into a separate PR and discussed on-list in a dedicated thread. I'm not very keen on seeing the inner/outer signature flipped without strong rationale because it will be a breaking change for the existing ACME client implementations.

Following up on my own comment: the inner/outer signature is not being flipped, the change is to the inner key parameter of the inner JWS' authenticated body. I have less opposition to that change but I think it would be helpful to call out the motivation explicitly :+1:

felixfontein:
I don't know the motivation, but for me, it makes sense if you interpret the key change as a transfer of the account from the owner of the old private key to the owner of the new private key. (Of course, this is only interesting if these are different entities.) With the old method, the owner of the new private key has to provide a signed message and his public key info to the owner of the old private key without knowing which account will be changed to her private key. With the new method, this is no longer possible.

draft-ietf-acme-acme.md
bifurcation: Ah, English...

bifurcation:
Yeah, the motivation is that the "handshake" is supposed to be symmetrical here:

* Old key signs over new key and account URL
* New key signs over old key and account URL

That way, both halves of the hand-off are represented explicitly.  In addition, the existing "newKey" field doesn't add any value, since the new key is already provided in the "jwk" field of the protected header.

Issue #431, opened 2018-08-08
Make "URL" more descriptive


bifurcation:
Jim Schaad's designated expert review suggested a more descriptive title than "URL" for the "url" JWS attribute.  For example, "Target URL".

bifurcation:
Fixed by #432

PR #432, opened 2018-08-09
spt's comments

seanturner:
Some of the nits I noted during my review.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -51,16 +51,23 @@ normative:
       name: Juraj Somorovsky
       ins: J. Somorovsky
     target: https://dl.acm.org/citation.cfm?id=2813657
+  REST:
+    title: Architectural Styles and the Design of Network-based Software Architectures
+    author:
+      name: Roy Thomas Fielding
+      ins: R. Fielding
+    date: 2000
+    target: http://www.ics.uci.edu/~fielding/pubs/dissertation/top.htm
 
 
 --- abstract
 
-Certificates in PKI using X.509 (PKIX) are used for a number of purposes,
+Public Key Infrastructure X.509 (PKIX) certificates are used for a number of purposes,
 the most significant of which is the authentication of domain names.  Thus,
-certificate authorities in the Web PKI are trusted to verify that an applicant
+certification authorities (CAs) in the Web PKI are trusted to verify that an applicant
 for a certificate legitimately represents the domain name(s) in the certificate.
 Today, this verification is done through a collection of ad hoc mechanisms.
-This document describes a protocol that a certification authority (CA) and an
+This document describes a protocol that a CA and an
 applicant can use to automate the process of verification and certificate
 issuance.  The protocol also provides facilities for other certificate
 management functions, such as certificate revocation.
@@ -76,7 +83,7 @@ discussed on the ACME mailing list (acme@ietf.org).
 # Introduction
 
 Certificates {{!RFC5280}} in the Web PKI are most commonly used to authenticate
-domain names.  Thus, certificate authorities in the Web PKI are trusted to
+domain names.  Thus, certification authorities (CAs) in the Web PKI are trusted to
 verify that an applicant for a certificate legitimately represents the domain
 name(s) in the certificate.
 
@@ -95,7 +102,7 @@ protocols for certificate issuance and identity verification.  In the case of DV
 certificates, a typical user experience is something like:
 
 * Generate a PKCS#10 {{!RFC2986}} Certificate Signing Request (CSR).
-* Cut-and-paste the CSR into a CA web page.
+* Cut-and-paste the CSR into a CA's web page.
 * Prove ownership of the domain by one of the following methods:
    * Put a CA-provided challenge at a specific place on the web server.
    * Put a CA-provided challenge in a DNS record corresponding to the target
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -51,16 +51,23 @@ normative:
...
+    date: 2000
+    target: http://www.ics.uci.edu/~fielding/pubs/dissertation/top.htm
 
 
 --- abstract
 
-Certificates in PKI using X.509 (PKIX) are used for a number of purposes,
+Public Key Infrastructure X.509 (PKIX) certificates are used for a number of purposes,
 the most significant of which is the authentication of domain names.  Thus,
-certificate authorities in the Web PKI are trusted to verify that an applicant
```
cpu: I'm a :-1: on changing all instances of "Certificate authority" to "Certification Authority". The later is not the predominant definition that folks are familiar with.

draft-ietf-acme-acme.md
```diff
@@ -395,8 +402,8 @@ An ACME server MUST implement the "ES256" signature algorithm {{!RFC7518}} and S
 The "jwk" and "kid" fields are mutually exclusive. Servers MUST reject requests
 that contain both.
 
-For newAccount requests, and for revokeCert requests authenticated by certificate
-key, there MUST be a "jwk" field. This field MUST contain the public key
+For newAccount requests, and for revokeCert requests authenticated by certified
+keys, there MUST be a "jwk" field. This field MUST contain the public key
```
cpu: I don't think changing "certificate key" to "certified keys" makes sense in this context. As a concrete example the account key that issued the certificate specified in the revocation request would be a certified key (in the sense it is authorized to do the revocation), but it wouldn't be a revocation request that MUST have a "jwk" field (it would carry the mutually exclusive "kid" field). 

This is specifically talking about the case where a request to revoke an issued end-entity certificate is authenticated using a JWK corresponding to the to-be-revoked certificate's public key.

draft-ietf-acme-acme.md
```diff
@@ -734,12 +742,12 @@ the following table and whose values are the corresponding URLs.
 
 | Field          | URL in value         |
 |:---------------|:---------------------|
-| newNonce       | New nonce            |
-| newAccount     | New account          |
-| newOrder       | New order            |
-| newAuthz       | New authorization    |
-| revokeCert     | Revoke certificate   |
-| keyChange      | Key change           |
+| newNonce       | new-nonce            |
```
cpu: I don't think these should be made to match example URLs. To me the table is meant to convey the fields (keys in the directory), and the human description of what endpoint URL is expected there, not an actual URL or reference to an example URL.

richsalz:
I think Certification Authority is the more-correct term.

cpu:
> I think Certification Authority is the more-correct term.

I'm advocating using the term that readers will be familiar with and understand not the term that is more-correct to domain experts within the IETF.

draft-ietf-acme-acme.md
seanturner: I'm certainly not planning on dying on this hill, but there there is some precedent to using it based on RFC 2459, 328, and 5280 as well as slew of other RFCs.  And, RFC 4949 says "certificate authority" SHOULD NOT be used in Standards Track documents.

draft-ietf-acme-acme.md
seanturner: Ah so maybe it's just missing "a" in the sentence "and for revokeCert requests authenticated by a certificate key".

richsalz:
The folks who are reading IETF RFC's are mostly IETF folks.  Consistency within the RFC's is important. Most people know the CA term, not what it stands for :)

cpu:
> The folks who are reading IETF RFC's are mostly IETF folks. 

Disagree. The vast majority of developers that have implemented the ACME draft so far are not people I see at IETF meetings or on the mailing lists.

draft-ietf-acme-acme.md
cpu: @seanturner :+1: I think adding a "a" in the sentence does clarify. Good suggestion

draft-ietf-acme-acme.md
cpu: Ok. I'll cave on this but I think its unfortunate that we're avoiding a term with common industry wide usage to favour one that is seemingly only favoured by folks already steeped in the nitty gritty of these standards who likely need the least help understanding an RFC and its contents.

bifurcation: I note that the CABF Baseline Requirements also use Certification Authority.  I agree that it's unpleasant, but it seems to be proper.

cpu:
@richsalz I replied again in my review comment to @seanturner: https://github.com/ietf-wg-acme/acme/pull/432/review_comment/create#discussion_r208973919 My appetite for bikeshedding is very low. There's clearly precedent to use "certification authority" in IETF land and I want to get back to work. I can live with `s/certificate authority/certification authority/g`.

seanturner:
Also closes #431.

draft-ietf-acme-acme.md
cpu: @seanturner Ping on this?

draft-ietf-acme-acme.md
```diff
@@ -453,7 +460,7 @@ same string in their "url" header parameter.
 The "url" header parameter specifies the URL {{!RFC3986}} to which this JWS
 object is directed.  The "url" header parameter MUST be carried in the protected
 header of the JWS.  The value of the "url" header parameter MUST be a string
-representing the URL.
+representing the Target URL.
```
cpu: I think this should probably be "target URL"

seanturner:
@cpu I think that does it.

draft-ietf-acme-acme.md
```diff
@@ -395,8 +402,8 @@ An ACME server MUST implement the "ES256" signature algorithm {{!RFC7518}} and S
 The "jwk" and "kid" fields are mutually exclusive. Servers MUST reject requests
 that contain both.
 
-For newAccount requests, and for revokeCert requests authenticated by certificate
-key, there MUST be a "jwk" field. This field MUST contain the public key
+For newAccount requests, and for revokeCert requests authenticated by a certificate
+keys, there MUST be a "jwk" field. This field MUST contain the public key
```
cpu: @seanturner I think "keys" should be singular now: "a certificate key". Sorry to flag one last thing.

PR #439, opened 2018-08-09
adding ref for Access-Control-Allow-Origin

### LINKED ISSUE #434: Good reference for Access-Control-Allow-Origin?
We need a normative reference for Access-Control-Allow-Origin.  https://www.iana.org/assignments/message-headers/message-headers.xhtml lists Access-Control-Allow-Origin as provisional.

seanturner:
@bifurcation Closes #434.  But, this might be kind of lame as the reference to CORS is in the preceding sentence.


richsalz:
I would close this and #434 as don't fix since it looks silly to repeat the ref in each sentence. In each sentence. :)

seanturner:
As I was entering the PR, I started to come around.

PR #441, opened 2018-08-29
Comments from Ben Campbell's IESG review

bifurcation:
https://datatracker.ietf.org/doc/draft-ietf-acme-acme/ballot/#ben-campbell
https://mailarchive.ietf.org/arch/msg/acme/ey499Irdc9VaW4-JG2-0A_HiwZs


bacampbe:
Looks good to me. I didn't necessarily expect the bit about de-provisioning a challenge response to need a SHOULD, but I don't object.

bifurcation:
Thanks @bacampbe.  With approval from co-author,  chair, and AD, I guess it's safe to merge :)

PR #442, opened 2018-08-29
Comments from Alexey Melnikov's IESG review

bifurcation:
https://datatracker.ietf.org/doc/draft-ietf-acme-acme/ballot/#alexey-melnikov
https://mailarchive.ietf.org/arch/msg/acme/_CQgGVs9pglD3ZMFg_MIPUJAKqQ

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -508,7 +508,7 @@ ignore invalid Replay-Nonce values.  The ABNF {{!RFC5234}} for the Replay-Nonce
 header field follows:
 
 ~~~~~
-  base64url = (%x41-5A) / (%x61-7A) / (%x30-39) / "-" / "_"
+  base64url = ALPHA / DIGIT / “-“ / “_”
 
   Replay-Nonce = 1*base64url
 ~~~~~
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -508,7 +508,7 @@ ignore invalid Replay-Nonce values.  The ABNF {{!RFC5234}} for the Replay-Nonce
 header field follows:
 
 ~~~~~
-  base64url = ALPHA / DIGIT / “-“ / “_”
+  base64url = ALPHA / DIGIT / "-" / "_"
 
   Replay-Nonce = 1*base64url
 ~~~~~
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -2489,7 +2489,9 @@ domain by verifying that the resource was provisioned as expected.
 5. Verify that key authorization provided by the HTTP server matches the key
    authorization stored by the server.
 
-The server SHOULD follow redirects when dereferencing the URL.  See
+The server SHOULD follow redirects when dereferencing the URL.
+Clients might use redirects, for example, so that the response can
+be provided by a centralized certificate management server.  See
 {{integrity-of-authorizations}} for security considerations related
 to redirects.
 
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -508,7 +508,7 @@ ignore invalid Replay-Nonce values.  The ABNF {{!RFC5234}} for the Replay-Nonce
 header field follows:
 
 ~~~~~
-  base64url = (%x40-5A) / (%x61-7A) / (%x30-39) / "-" / "_"
+  base64url = (%x41-5A) / (%x61-7A) / (%x30-39) / "-" / "_"
 
   Replay-Nonce = 1*base64url
 ~~~~~
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -783,8 +783,9 @@ caaIdentities (optional, array of string):
 itself for the purposes of CAA record validation as defined in
 {{!RFC6844}}.  Each string MUST represent the same sequence of ASCII
 code points that the server will expect to see as the "Issuer Domain
-Name.  This allows clients to determine the correct issuer domain
-name to use when configuring CAA records.
+Name" in a CAA issue or issuewild property tag.  This allows clients
+to determine the correct issuer domain name to use when configuring
+CAA records.
 
 externalAccountRequired (optional, boolean):
 : If this field is present and set to "true", then the CA requires that all
@@ -828,7 +829,7 @@ contact (optional, array of string):
 : An array of URLs that the server can use to contact the client for issues
 related to this account. For example, the server may wish to notify the
 client about server-initiated revocation or certificate expiration.
-For infomration on supported URL schemes, see {{account-creation}}
+For information on supported URL schemes, see {{account-creation}}
 
 termsOfServiceAgreed (optional, boolean):
 : Including this field in a new-account request, with a value of true, indicates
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -779,9 +779,11 @@ website (optional, string):
 information about the ACME server.
 
 caaIdentities (optional, array of string):
-: Each string MUST be a lowercase hostname which the ACME server recognizes as
-referring to itself for the purposes of CAA record validation as defined in
-{{!RFC6844}}.  This allows clients to determine the correct issuer domain name to
+: The hostnames that the ACME server recognizes as referring to
+itself for the purposes of CAA record validation as defined in
+{{!RFC6844}}.  Each string MUST represent the same sequence of ASCII
+code points that the server will expect to see in a CAA record.
```
cpu: I think we should borrow RFC6844's terminology and be a little bit more precise here. Something like:
"Each string MUST represent the same sequence of ASCII code points that the server will expect to see as the "Issuer Domain Name" in a CAA issue or issuewild property tag"

draft-ietf-acme-acme.md
```diff
@@ -2572,6 +2574,14 @@ checks, then the validation fails.
 
 ## MIME Type: application/pem-certificate-chain
 
+A file of this type contains one or more certificates encoded with the PEM textual encoding, according to
+RFC 7468 {{!RFC7468}}.  In order to provide easy interoperation with TLS, the first
+certificate MUST be an end-entity certificate. Each following certificate
+SHOULD directly certify the one preceding it. Because certificate validation
+requires that trust anchors be distributed independently, a certificate
+that specifies a trust anchor MAY be omitted from the chain, provided
```
cpu: This says "a certificate that specifies a trust anchor MAY be omitted from the chain" but that makes it sound like the certificate whose issuer is a trust anchor specified out of band can be omitted when I think the intention is that the trust anchor itself may be omitted.

Can you clarify this? Maybe:
> Because certificate validation requires that trust anchors be distributed independently required trust anchors MAY be omitted from the chain, provided that supported peers are known to possess the omitted trust anchor certificates

(or similar?)

draft-ietf-acme-acme.md
```diff
@@ -2876,7 +2894,7 @@ Label column of the ACME Identifier Types registry.
 Template:
 
 * Label: The identifier for this validation method
-* Identifier Type: The type of identifier that this method applies to
+* Identifier Type: The type of identifier that this method applies to.  For non-ACME validation methods, this field should be set to "N/A".
```
cpu: Should non-ACME validation methods require a non-ACME identifier type? I thought you would be using non-ACME validation methods to validate DNS type identifiers.

draft-ietf-acme-acme.md
bifurcation: 👍

draft-ietf-acme-acme.md
bifurcation: I didn't change the text here, just copy/pasted it from below :)  Nonetheless, happy to use your text.

bifurcation: I'm just going to change "specifies" to "represents".  I think that addresses the ambiguity you note; it doesn't seem like a cert under a TA would be understood to represent that TA.

cpu: > I didn't change the text here, just copy/pasted it from below

:laughing: Doh!

> I'm just going to change "specifies" to "represents". I think that addresses the ambiguity you note; it doesn't seem like a cert under a TA would be understood to represent that TA.

That works for me :+1: Thanks

draft-ietf-acme-acme.md
bifurcation: I guess that's true in principle, but you're not doing it **in ACME**, so it doesn't seem crucial to specify an ACME identifier type.  That's probably parsing things too closely, though.  I'll change back to "RESERVED".

draft-ietf-acme-acme.md
```diff
@@ -779,10 +779,12 @@ website (optional, string):
...
 caaIdentities (optional, array of string):
-: Each string MUST be a lowercase hostname which the ACME server recognizes as
-referring to itself for the purposes of CAA record validation as defined in
-{{!RFC6844}}.  This allows clients to determine the correct issuer domain name to
-use when configuring CAA records.
+: The hostnames that the ACME server recognizes as referring to
+itself for the purposes of CAA record validation as defined in
+{{!RFC6844}}.  Each string MUST represent the same sequence of ASCII
+code points that the server will expect to see as the "Issuer Domain
+Name.  This allows clients to determine the correct issuer domain
```
cpu: You lost a closing `"` in this. Did you want to avoid the reference to the specific "issue/issuewild property tag"? I thought it was useful but I'm not attached if you disagreed.

draft-ietf-acme-acme.md
```diff
@@ -826,6 +828,7 @@ contact (optional, array of string):
 : An array of URLs that the server can use to contact the client for issues
 related to this account. For example, the server may wish to notify the
 client about server-initiated revocation or certificate expiration.
+For infomration on supported URL schemes, see {{account-creation}}
```
cpu: typo: "infomration" -> "information"

draft-ietf-acme-acme.md
```diff
@@ -2876,11 +2895,15 @@ Label column of the ACME Identifier Types registry.
 Template:
 
 * Label: The identifier for this validation method
-* Identifier Type: The type of identifier that this method applies to
+* Identifier Type: The type of identifier that this method applies to.
```
cpu: This template section is inconsistent with respect to the trailing `.` - "Identifier Type" and "ACME" use a trailing period, the "Label" and "Reference" bullets do not. I don't have a preference except that they be consistent :-)

draft-ietf-acme-acme.md
```diff
@@ -2876,11 +2895,15 @@ Label column of the ACME Identifier Types registry.
...
 * ACME: "Y" if the validation method corresponds to an ACME challenge type;
   "N" otherwise.
 * Reference: Where the validation method is defined
 
+This registry may also contain reserved entries (e.g., to avoid
+collisions).  Such entries should have the"ACME" field set to "N"
```
cpu: typo: `the"ACME"` -> `the "ACME"`.

draft-ietf-acme-acme.md
bifurcation: Nope, just didn't copy/paste properly 🙄

draft-ietf-acme-acme.md
```diff
@@ -1939,7 +1943,7 @@ own heuristics to decide which is optimal.
 ~~~~~~~~~~
 GET /acme/cert/asdf HTTP/1.1
 Host: example.com
-Accept: application/pkix-cert
+Accept: application/pem-certificate-chain
```
kaduk: Couldn't we list both?

draft-ietf-acme-acme.md
```diff
@@ -2485,7 +2489,9 @@ domain by verifying that the resource was provisioned as expected.
 5. Verify that key authorization provided by the HTTP server matches the key
    authorization stored by the server.
 
-The server SHOULD follow redirects when dereferencing the URL.
+The server SHOULD follow redirects when dereferencing the URL.  See
+{{integrity-of-authorizations}} for security considerations related
+to redirects.
```
kaduk: Perhaps this is getting into my ballot remark instead of Alexey's, but the security considerations mostly tell you about risks of following the redirect; do we want to say anything about the advantages/which cases it is useful for?

draft-ietf-acme-acme.md
```diff
@@ -2583,26 +2597,34 @@ Required parameters: None
...
-SHOULD directly certify the one preceding it. Because certificate validation
-requires that trust anchors be distributed independently, a certificate
-that specifies a trust anchor MAY be omitted from the chain, provided
-that supported peers are known to possess any omitted certificates.
+  Deprecated alias names for this type: n/a
+  Magic number(s): n/a
+  File extension(s): .pem
+  Macintosh file type code(s): n/a
+
+Person & email address to contact for further information: See Authors' Addresses section.
```
kaduk: IANA and the RFC Editor should see that section, right, not the consumer of the registry?

draft-ietf-acme-acme.md
bifurcation: We could, but I think that would just add ambiguity to the example.

draft-ietf-acme-acme.md
bifurcation: AFAICT, this information doesn't go in the registry anyway, it just lives in the RFC.

PR #444, opened 2018-08-30
Adjusting example to match new text about potentially sensitive resources

adamroach:


bifurcation:
This is obsoleted by #445

PR #445, opened 2018-08-30
Remove GET requests

bifurcation:
This is one option for resolving the privacy concerns raised in Adam's and Ben's AD reviews.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -469,7 +469,7 @@ process, e.g., the web server that will use the referenced
 certificate chain.  (See {{?I-D.ietf-acme-star}} for more advanced
 cases.)  A server that allows GET requests for certificate resources
 can still provide a degree of access control by assigning them
-capability URLs them capability URLs {{?W3C.WD-capability-urls-20140218}}.
+capability URLs {{?W3C.WD-capability-urls-20140218}}.
 As above, if the server does not allow GET requests for a given
 resource, it MUST return an error with status code 405 "Method Not
 Allowed" and type "malformedRequest".


draft-ietf-acme-acme.md
```diff
@@ -411,10 +412,25 @@ For all other requests, the request is signed using an existing account and
...
+MUST allow GET requests for its directory resource and both GET and
+HEAD requests for its newNonce resource (see {{resources}}), in
+order to support client bootstrapping.
+
+Otherwise, if a client wishes to fetch a resource from
+the server (which would otherwise be done with a GET), then it MUST
+send a POST request with a JWS body as described above, with the
+following differences:
+
+* The JWS Protected Header MUST contain a "typ" field with the value
```
felixfontein: Shouldn't it be "type", not "typ"?

draft-ietf-acme-acme.md
```diff
@@ -790,7 +806,7 @@ externalAccountRequired (optional, boolean):
 new-account requests include an "externalAccountBinding" field associating the
 new account with an external account.
 
-Clients access the directory by sending a GET request to the directory URL.
+Clients access the directory by sending a POST-as-GET request to the directory URL.
```
felixfontein: This should be `GET`, as this is about the directory (which is explicitly mentioned as `GET` above).

draft-ietf-acme-acme.md
felixfontein: I noticed that there exists a JWS header called ["typ"](https://tools.ietf.org/html/rfc7515#section-4.1.9), but that's supposed to be a media type. Is it ok to redefine "typ" for ACME?

jsha:
I think we don't need to require a "typ" field. I think we just need to update the semantics for each resource so that POSTs are always allowed, and a POST with the literal contents `{}` is a request to echo the current contents of that resource. This is what we already do for retrieving account objects, so we should mirror that.

draft-ietf-acme-acme.md
kaduk: https://tools.ietf.org/html/rfc7515#section-4.1.9

draft-ietf-acme-acme.md
```diff
@@ -411,10 +412,25 @@ For all other requests, the request is signed using an existing account and
...
+HEAD requests for its newNonce resource (see {{resources}}), in
+order to support client bootstrapping.
+
+Otherwise, if a client wishes to fetch a resource from
+the server (which would otherwise be done with a GET), then it MUST
+send a POST request with a JWS body as described above, with the
+following differences:
+
+* The JWS Protected Header MUST contain a "typ" field with the value
+  "GET"
```
kaduk: Don't we need IANA considerations to register this header parameter?

draft-ietf-acme-acme.md
```diff
@@ -720,18 +736,18 @@ establish a new account with the server, prove control of an identifier, issue a
...
+| Get directory         | GET  directory             | 200            |
+| Get nonce             | HEAD newNonce              | 200            |
+| Create account        | POST newAccount            | 201 -> account |
+| Submit order          | POST newOrder              | 201 -> order   |
+| Fetch challenges      | POST-as-GET authorizations | 200            |
+| Respond to challenges | POST challenge urls        | 200            |
+| Poll for status       | POST-as-GET order          | 200            |
+| Finalize order        | POST order finalize        | 200            |
+| Poll for status       | POST-as-GET order          | 200            |
+| Download certificate  | POST order certificate     | 200            |
```
kaduk: I'm feeling particularly dense today (jet lag), but why is this POST and not POST-as-GET?

draft-ietf-acme-acme.md
```diff
@@ -1938,10 +1946,23 @@ paths to various trust anchors. Clients can fetch these alternates and use their
...
 Host: example.com
+Content-Type: application/jose+json
 Accept: application/pkix-cert
 
+{
+  "protected": base64url({
+    "alg": "ES256",
+    "kid": "https://example.com/acme/acct/1",
+    "nonce": "uQpSjlRb4vQVCjVYAyyUWg",
+    "url": "https://example.com/acme/new-authz",
```
cpu: This `"url"` header value is not correct for a POST to `/acme/cert/asdf`

draft-ietf-acme-acme.md
```diff
@@ -1996,14 +2017,28 @@ authorization object is implicitly tied to the account key used to sign the
...
 Host: example.com
+Content-Type: application/jose+json
+Accept: application/pkix-cert
+
+{
+  "protected": base64url({
+    "alg": "ES256",
+    "kid": "https://example.com/acme/acct/1",
+    "nonce": "uQpSjlRb4vQVCjVYAyyUWg",
+    "url": "https://example.com/acme/new-authz",
```
cpu: This `"url"` header value is not correct for a POST to `/acme/authz/1234`

draft-ietf-acme-acme.md
```diff
@@ -2093,15 +2128,29 @@ where the client can tell when the server has validated the challenge (e.g., by
...
```
cpu: This "url" header value is not correct for a POST to /acme/authz/1234

cpu:
> I think we don't need to require a "typ" field. I think we just need to update the semantics for each resource so that POSTs are always allowed, and a POST with the literal contents {} is a request to echo the current contents of that resource. This is what we already do for retrieving account objects, so we should mirror that.

@jsha the POST to initiate a challenge throws a :wrench: into this plan. Specifically the server needs a way to differentiate a POST to an authorization's challenge as a POST to an authorization's challenge to initiate the challenge validation process or a POST-AS-GET to poll the challenge. Both types of POST would be using a `{}` body and without the `typ` header or equivalent I don't think it would be possible to know which was intended.

draft-ietf-acme-acme.md
```diff
@@ -411,10 +412,25 @@ For all other requests, the request is signed using an existing account and
...
+order to support client bootstrapping.
+
+Otherwise, if a client wishes to fetch a resource from
+the server (which would otherwise be done with a GET), then it MUST
+send a POST request with a JWS body as described above, with the
+following differences:
+
+* The JWS Protected Header MUST contain a "typ" field with the value
+  "GET"
+* The payload MUST be an empty JSON object ({})
```
cpu: What should the server return to the client if this isn't true? A `urn:ietf:params:acme:error:unauthorized` type problem? `urn:ietf:acme:error:malformedRequest`?

draft-ietf-acme-acme.md
```diff
@@ -411,10 +412,25 @@ For all other requests, the request is signed using an existing account and
...
+
+We will refer to these as "POST-as-GET" requests. On receiving such
+a request, the server MUST authenticate the sender and verify any
+access control rules.  Otherwise, the server MUST treat this request
```
cpu: What should the server return to the client if the POST-as-GET authentication fails? A urn:ietf:params:acme:error:unauthorized type problem? urn:ietf:acme:error:malformedRequest?

draft-ietf-acme-acme.md
```diff
@@ -411,10 +412,25 @@ For all other requests, the request is signed using an existing account and
...
+Note that authentication via signed JWS request bodies implies that
+requests other than POST requests are not authenticated.  The server
+MUST allow GET requests for its directory resource and both GET and
+HEAD requests for its newNonce resource (see {{resources}}), in
+order to support client bootstrapping.
+
+Otherwise, if a client wishes to fetch a resource from
+the server (which would otherwise be done with a GET), then it MUST
+send a POST request with a JWS body as described above, with the
+following differences:
```
cpu: What should the server return if it receives a GET to a resource? SHOULD (MUST?) return a `urn:ietf:acme:error:malformedRequest` problem with status 405?

draft-ietf-acme-acme.md
bifurcation: @adamroach pointed out to me that the `typ` header parameter is tied to media types, though, so we'll have to pick a different signal here.  Because `application/GET` isn't a thing.

draft-ietf-acme-acme.md
bifurcation: Yep.  Overzealous search-and-replace :)

draft-ietf-acme-acme.md
bifurcation: No; in fact, I chose `typ` because it was already registered.  However, as noted above, that'll need to change.

draft-ietf-acme-acme.md
```diff
@@ -429,6 +424,42 @@ set to "application/jose+json".  If a request does not meet this
...
+a JWS body as described above, where the payload of the JWS is a
+zero-length octet string.
+
+We will refer to these as "POST-as-GET" requests. On receiving a
+request with a zero-length (and thus non-JSON) payload, the server
+MUST authenticate the sender and verify any access control rules.
+Otherwise, the server MUST treat this request as having the same
+semantics as a GET request for the same resource.
+
+The server MUST allow GET requests for the following resources(see
```
cpu: typo: `resources(see)` -> `resources (see)`

draft-ietf-acme-acme.md
```diff
@@ -429,6 +424,42 @@ set to "application/jose+json".  If a request does not meet this
...
+{{resources}}, in addition to POST-as-GET requests for these
+resources:
```
cpu: nit: I think the second "these resources" in this sentence is redundant.

draft-ietf-acme-acme.md
```diff
@@ -429,6 +424,42 @@ set to "application/jose+json".  If a request does not meet this
...
+
+* The directory resource
+* The newNonce resource
+* Certificate resources
+
+Allowing GET requests for the directory and newNonce resourcse
```
cpu: typo: "resourcse" -> "resource"

draft-ietf-acme-acme.md
```diff
@@ -1616,7 +1639,9 @@ orders or authorization transactions based on a change of account key.
 
 A client can deactivate an account by posting a signed update to the server with
 a status field of "deactivated." Clients may wish to do this when the account
-key is compromised or decommissioned.
+key is compromised or decommissioned. A deactivated account an no longer request
```
cpu: typo: "an no longer" -> "can no longer"

draft-ietf-acme-acme.md
```diff
@@ -1996,14 +2020,28 @@ authorization object is implicitly tied to the account key used to sign the
...
+
+{
+  "protected": base64url({
+    "alg": "ES256",
+    "kid": "https://example.com/acme/acct/1",
+    "nonce": "uQpSjlRb4vQVCjVYAyyUWg",
+    "url": "https://example.com/acme/authz/1234",
+    "typ": "GET"
+  }),
+  "payload": base64url({}),
```
cpu: Shouldn't this be `"payload": null,` now since its replacing a GET to this authz with a POST-as-GET?

draft-ietf-acme-acme.md
```diff
@@ -2093,15 +2131,29 @@ where the client can tell when the server has validated the challenge (e.g., by
...
```
cpu: ditto here RE: the null payload.

draft-ietf-acme-acme.md
jsha: I think this should actually be `"payload": ""`, since JSON distinguishes between an empty string (or equivalently, zero-length payload) and the null value.

draft-ietf-acme-acme.md
kaduk: Somehow I thought we had a registry of values that the "typ" field could take.  Which, I guess we do, if it's media types, but I was clearly confused about it.

draft-ietf-acme-acme.md
```diff
@@ -443,22 +443,19 @@ MUST authenticate the sender and verify any access control rules.
...
+for these resources.  This enables clients to bootstrap into the
+ACME authentication system.
+
+The server MAY allow GET requests for certificate resources, in
+order to allow certificates to be fetched by a lower-privileged
+process, e.g., the web server that will use the referenced
+certificate chain.  (See {{?I-D.ietf-acme-star}} for more advanced
+cases.)  A server that allows GET requests for certificate resources
+can still provide them a degree of access control by assigning them
+capability URLs them capability URLs
```
felixfontein: "them capability URLs" is there twice.

yaronf: Also please add a reference for "capability URLs", https://www.w3.org/TR/capability-urls/

draft-ietf-acme-acme.md
```diff
@@ -724,18 +752,18 @@ establish a new account with the server, prove control of an identifier, issue a
...
+| Get directory         | GET  directory             | 200            |
+| Get nonce             | HEAD newNonce              | 200            |
+| Create account        | POST newAccount            | 201 -> account |
+| Submit order          | POST newOrder              | 201 -> order   |
+| Fetch challenges      | POST-as-GET authorizations | 200            |
+| Respond to challenges | POST challenge urls        | 200            |
+| Poll for status       | POST-as-GET order          | 200            |
+| Finalize order        | POST order finalize        | 200            |
+| Poll for status       | POST-as-GET order          | 200            |
+| Download certificate  | GET  order certificate     | 200            |
```
yaronf: Should the "Download certificate" row be "GET or POST-as-GET"?

draft-ietf-acme-acme.md
```diff
@@ -3233,6 +3271,30 @@ CAs using ACME to allow clients to agree to terms of service should keep in mind
...
+ACME does not specify how the server constructs the URLs that it
+uses to address resources.  If the server operator uses URLs that
+are predictable to third parties, this can leak information about
+what URLs exist on the server, since an attacker can probe for
+whether POST-as-GET request to the URL returns "Not Found" or
+"Unauthorized".  
+
+For example, suppose that the CA uses highly structured URLs that
+with several low-entropy fields:
+
```
yaronf: The "For example" sentence above is incomplete.

draft-ietf-acme-acme.md
```diff
@@ -1821,7 +1843,7 @@ action the client should take:
   fulfilled, and is awaiting finalization.  Submit a finalization
   request.
 
-* "processing": The certificate is being issued. Send a GET request after the
+* "processing": The certificate is being issued. Send a POST-as-GET request after the
```
yaronf: This request is for the actual certificate, so should be GET or POST-as-GET.

draft-ietf-acme-acme.md
yaronf: Suggest to add: The server MAY choose to allow GET requests to certain certificate resources but not to others. The server can base this decision on out-of-band knowledge (e.g., to allow GET requests to certificates owned by a certain account) or on order-specific information, such as the extension defined in {{?I-D.ietf-acme-star}}.

draft-ietf-acme-acme.md
bifurcation: My thinking here is to have the description focus on the more secure case (which should be the default for most clients), and let the MAY for GETs stand alone above.

yaronf:
I was referring to the "Downloading the Certificate" section. After 
rereading the whole thing again, I realize that GET is still allowed. 
But still, for clarity, we should mention this fact again in this section.

richsalz:
I think we need to give the WG a chance to object to this.  Yes there has been some discussion, but I'll do a formal 'call' email.  Don't let that delay merging, we can always revert. :)

draft-ietf-acme-acme.md
```diff
@@ -433,6 +428,42 @@ set to "application/jose+json".  If a request does not meet this
...
+MUST authenticate the sender and verify any access control rules.
+Otherwise, the server MUST treat this request as having the same
+semantics as a GET request for the same resource.
+
+The server MUST allow GET requests for the directory and newNonce
+resources (see {{resources}}), in addition to POST-as-GET requests
+for these resources.  This enables clients to bootstrap into the
+ACME authentication system.
+
+The server MAY allow GET requests for certificate resources, in
```
cpu: minor nit: I don't think the comma after "resources" is required here.

draft-ietf-acme-acme.md
```diff
@@ -433,6 +428,42 @@ set to "application/jose+json".  If a request does not meet this
...
+order to allow certificates to be fetched by a lower-privileged
+process, e.g., the web server that will use the referenced
+certificate chain.  (See {{?I-D.ietf-acme-star}} for more advanced
+cases.)  A server that allows GET requests for certificate resources
+can still provide them a degree of access control by assigning them
```
cpu: nit: "can still provide them a degree" -> "can still provide a degree"

draft-ietf-acme-acme.md
```diff
@@ -1624,7 +1647,9 @@ orders or authorization transactions based on a change of account key.
 
 A client can deactivate an account by posting a signed update to the server with
 a status field of "deactivated." Clients may wish to do this when the account
-key is compromised or decommissioned.
+key is compromised or decommissioned. A deactivated account can no longer request
+certificate issuance or access resources related to the account, such as orders
+or authorizations.
```
cpu: Can you add one more sentence that explicitly says "If a server receives a POST or POST-as-GETs from a deactivated account it MUST return a urn:ietf:params:acme:error:unauthorized problem"?

draft-ietf-acme-acme.md
```diff
@@ -453,12 +453,12 @@ resources (see {{resources}}), in addition to POST-as-GET requests
...
 
-The server MAY allow GET requests for certificate resources, in
+The server MAY allow GET requests for certificate resources in
 order to allow certificates to be fetched by a lower-privileged
 process, e.g., the web server that will use the referenced
 certificate chain.  (See {{?I-D.ietf-acme-star}} for more advanced
 cases.)  A server that allows GET requests for certificate resources
-can still provide them a degree of access control by assigning them
+can still provide a degree of access control by assigning them
 capability URLs them capability URLs {{?W3C.WD-capability-urls-20140218}}.
```
kaduk: "them capability URLs" they really are a-happen-ing ... twice.

PR #447, opened 2018-08-31
Comments from Ben Kaduk's IESG review

bifurcation:
https://datatracker.ietf.org/doc/draft-ietf-acme-acme/ballot/#benjamin-kaduk
https://mailarchive.ietf.org/arch/msg/acme/YLYlBVdnv2VxllVCS8pF-QueLe8

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -354,7 +354,9 @@ validating an 'http-01' challenge, a DNS client with 'dns-01', etc.
 ACME servers SHOULD follow the recommendations of {{?RFC7525}} when configuring
 their TLS implementations.  ACME servers that support TLS 1.3 MAY allow clients
 to send early data (0-RTT).  This is safe because the ACME protocol itself
-includes anti-replay protections (see {{replay-protection}}).
+includes anti-replay protections (see {{replay-protection}}) in all cases where
+they are required.  For this reason, there are no restrictions on what ACME
+data can be carried in 0-RTT. 
 
 ACME clients MUST send a User-Agent header, in accordance with
 {{!RFC7231}}. This header SHOULD include the name and version of the
@@ -1017,11 +1019,13 @@ encoded in the format specified in RFC 3339 {{!RFC3339}}.  This field is REQUIRE
 for objects with "valid" in the "status" field.
 
 challenges (required, array of objects):
-: For pending authorizations, the challenges that the client can fulfill in
-order to prove possession of the identifier.  For final authorizations (in the "valid" or "invalid" state), the
-challenges that were used.  Each array entry is an object with parameters
-required to validate the challenge.  A client should attempt to fulfill
-one of these challenges, and a server should consider any one of the challenges
+: For pending authorizations, the challenges that the client can
+fulfill in order to prove possession of the identifier.  For valid
+authorizations, the challenge that was validated.  For invalid
+authorizations, the challenges that were attempted and failed.  Each
+array entry is an object with parameters required to validate the
+challenge.  A client should attempt to fulfill one of these
+challenges, and a server should consider any one of the challenges
 sufficient to make the authorization valid.
 
 wildcard (optional, boolean):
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -2982,6 +2986,12 @@ ensuring that an attacker with access to the validation channel can't obtain
 illegitimate authorization by acting as an ACME client (legitimately, in terms
 of the protocol).
 
+We do not protect against other types of abuse by a MitM on the ACME channel.
```
richsalz: s/We/ACME/ ?

draft-ietf-acme-acme.md
```diff
@@ -1017,11 +1019,13 @@ encoded in the format specified in RFC 3339 {{!RFC3339}}.  This field is REQUIRE
...
 challenges (required, array of objects):
-: For pending authorizations, the challenges that the client can fulfill in
-order to prove possession of the identifier.  For final authorizations (in the "valid" or "invalid" state), the
-challenges that were used.  Each array entry is an object with parameters
-required to validate the challenge.  A client should attempt to fulfill
-one of these challenges, and a server should consider any one of the challenges
+: For pending authorizations, the challenges that the client can
+fulfill in order to prove possession of the identifier.  For valid
+authorizations, the challenge that was validated.  For invalid
+authorizations, the challenges that were attempted and failed.  Each
```
kaduk: I suspect that the text for the "invalid" case could be interpreted multiple ways, but I'm also unsure that it really matters very much.  (That is, does it only include ones that the client responded to but the server failed to validate, or does it include all the challenges that were originally present in the "pending" case?)

draft-ietf-acme-acme.md
kaduk: Yeah, "ACME does not protect" is probably better.

draft-ietf-acme-acme.md
```diff
@@ -2982,6 +2986,12 @@ ensuring that an attacker with access to the validation channel can't obtain
 illegitimate authorization by acting as an ACME client (legitimately, in terms
 of the protocol).
 
+We do not protect against other types of abuse by a MitM on the ACME channel. 
+For example, such an attacker could send a bogus "badSignatureAlgorithm"
+response to downgrade a client to the lowest-quality signature algorithm that
+the server supports.  A MitM that is present on all connections (such as a
```
kaduk: If we end up concluding there is something interesting to say about caaIdentities, we could mention it here as well.

draft-ietf-acme-acme.md
bifurcation: I would expect it would include all the challenges in the "invalid" state, i.e., ones that the client tried and failed to validate.  But I agree that it doesn't really matter.

draft-ietf-acme-acme.md
bifurcation: I think this just falls under the denial-of-service heading.

draft-ietf-acme-acme.md
cpu: I think it should say "the challenge that was attempted and failed", singular "challenge" vs "challenges". The later text says: "A client should attempt to fulfill **one** of these challenges".

draft-ietf-acme-acme.md
```diff
@@ -2982,6 +2986,12 @@ ensuring that an attacker with access to the validation channel can't obtain
 illegitimate authorization by acting as an ACME client (legitimately, in terms
 of the protocol).
 
+ACME does not protect against other types of abuse by a MitM on the ACME channel. 
+For example, such an attacker could send a bogus "badSignatureAlgorithm"
+response to downgrade a client to the lowest-quality signature algorithm that
```
cpu: nit: "response" -> "error response"?

draft-ietf-acme-acme.md
bifurcation: I don't think there's anything to say that a server MUST fail the authz if a single challenge fails.

bifurcation: That is, I think it's valid behavior for the server to keep the authz open and allow the client to try multiple challenges until it fulfills one or the server gives up.  In other other words, I read the "one" as attaching to the "fulfill", not the "attempt".

cpu: We talked OOB about this and agree the specification is vague on this point.

I would be in favour of clarifying that an invalid challenge == an invalid authorization. This matches the Let's Encrypt implementation and when in doubt of the decision to make I'm firmly in favour of standardizing what has implementation experience in the client/server ecosystem today.

draft-ietf-acme-acme.md
```diff
@@ -1121,32 +1121,34 @@ valid              invalid
 Authorization objects are created in the "pending" state.  If one of
 the challenges listed in the authorization transitions to the
 "valid" state, then the authorization also changes to the "valid"
-state.  If there is an error while the authorization is still
+state.  If the client attemts to fulfill a challenge and failes, or
```
kaduk: nit: "fails"

PR #448, opened 2018-08-31
Responses to Adam Roach's IESG review COMMENTs

bifurcation:
https://datatracker.ietf.org/doc/draft-ietf-acme-acme/ballot/#benjamin-kaduk
https://mailarchive.ietf.org/arch/msg/acme/YLYlBVdnv2VxllVCS8pF-QueLe8

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -58,13 +58,6 @@ normative:
       ins: R. Fielding
     date: 2000
     target: http://www.ics.uci.edu/~fielding/pubs/dissertation/top.htm
-  ECMAScript:
-    title: ECMAScript 2018 Language Specification
-    author:
-      name: ECMA
-      ins: ECMA
-    date: 2018-06
-    target: http://www.ecma-international.org/ecma-262/
 
 
 --- abstract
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -322,12 +329,18 @@ The use of ACME for other identifiers will require further specification in orde
...
+# Encodings
 
 All requests and responses sent via HTTP by ACME clients, ACME servers, and
 validation servers as well as any inputs for digest computations MUST be encoded
 using the UTF-8 {{!RFC3629}} character set.
 
+Some JSON structures used by ACME include date time values encoded
+as strings.  These values MUST be encoded in the format specified in
+Section 15.0.1.15 of the ECMAScript specification {{ECMAScript}}.
+This format is a simplification of the format defined in
```
richsalz: s/simplification/profile/ maybe?

draft-ietf-acme-acme.md
```diff
@@ -322,12 +329,18 @@ The use of ACME for other identifiers will require further specification in orde
...
 validation challenges the server might require.
 
-# Character Encoding
+# Encodings
 
 All requests and responses sent via HTTP by ACME clients, ACME servers, and
 validation servers as well as any inputs for digest computations MUST be encoded
 using the UTF-8 {{!RFC3629}} character set.
 
+Some JSON structures used by ACME include date time values encoded
```
richsalz: s/date time/date-time/ ?
But it seems you say "date format" elsewhere.  Just do the right thing. :)

draft-ietf-acme-acme.md
```diff
@@ -889,7 +902,7 @@ status (required, string):
 
 expires (optional, string):
 : The timestamp after which the server will consider this order invalid, encoded
-in the format specified in RFC 3339 {{!RFC3339}}.  This field is REQUIRED for
+in the date format specified in {{encodings}}.  This field is REQUIRED for
```
richsalz: I'd drop the word "date" or maybe use "date-time" or maybe change to "time"

draft-ietf-acme-acme.md
bifurcation: I rolled this back.  It's incompatible with some other libraries (e.g., Go's `time`)

jsha: I don't think it's accurate that Go's `time` can't handle the date format defined in ECMAScript. Here's an example using a custom layout:

https://play.golang.org/p/6kTORVzXh5e

PR #454, opened 2018-10-02
Minor fixes related to POST-as-GET

bifurcation:
1. As suggested by David Hancock, change the certificate fetch examples to use POST-as-GET, noting that GET is optional

2. As suggested by Felix Fontein, use the correct error code for "malformed" in the description of what servers should do in response to inappropriate GET requests

PR #455, opened 2018-10-02
Randomize URLs in examples

bifurcation:
This PR updates the examples in the document to use the randomized URL scheme recommended in #445, in order to make clear the guidance to implementers.

Depends on #454

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -472,7 +472,7 @@ can still provide a degree of access control by assigning them
 capability URLs {{?W3C.WD-capability-urls-20140218}}.
 As above, if the server does not allow GET requests for a given
 resource, it MUST return an error with status code 405 "Method Not
-Allowed" and type "malformedRequest".
+Allowed" and type "malformed".
 
 ## Request URL Integrity
 
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -896,7 +896,7 @@ a POST-as-GET request, as described in {{orders-list}}.
     "mailto:admin@example.com"
   ],
   "termsOfServiceAgreed": true,
-  "orders": "https://example.com/acme/acct/1/orders"
+  "orders": "https://example.com/acme/acct/evOfKhNU60wg/orders"
 }
 ~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -2497,7 +2515,7 @@ on randomness requirements.
 ~~~~~~~~~~
 {
   "type": "http-01",
-  "url": "https://example.com/acme/authz/0",
+  "url": "https://example.com/acme/authz/eYL3Xw",
```
felixfontein: Is there a reason why `eYL3Xw` is much sohrter than the other authz identifiers used in the draft (`PAniVnsZcis`, `r4HqLzrSrpI`)?

draft-ietf-acme-acme.md
```diff
@@ -1995,9 +1995,21 @@ paths to various trust anchors. Clients can fetch these alternates and use their
...
-Accept: application/pem-certificate-chain
+Content-Type: application/jose+json
+Accept: application/pkix-cert
+
+{
+  "protected": base64url({
+    "alg": "ES256",
+    "kid": "https://example.com/acme/acct/evOfKhNU60wg",
+    "nonce": "uQpSjlRb4vQVCjVYAyyUWg",
+    "url": "https://example.com/acme/cert/mAt3xBGaobw",
```
felixfontein: The trailing comma is invalid JSON.

draft-ietf-acme-acme.md
```diff
@@ -2059,17 +2077,17 @@ will have already received the pending authorization object in the response
...
 Accept: application/pkix-cert
 
 {
   "protected": base64url({
     "alg": "ES256",
-    "kid": "https://example.com/acme/acct/1",
+    "kid": "https://example.com/acme/acct/evOfKhNU60wg",
     "nonce": "uQpSjlRb4vQVCjVYAyyUWg",
-    "url": "https://example.com/acme/authz/1234",
+    "url": "https://example.com/acme/authz/PAniVnsZcis",
```
felixfontein: The trailing comma is invalid JSON.

draft-ietf-acme-acme.md
```diff
@@ -2169,17 +2187,17 @@ progress, the server MUST return a 200 (OK) response and MAY include a
...
```
felixfontein: The trailing comma is invalid JSON.

draft-ietf-acme-acme.md
bifurcation: Search/replace error.  That should have been a challenge URL as well.

jsha:
I oppose this PR, because I oppose the advice added in #445 to use capability URLs, since it dilutes our authentication model unnecessarily. Sorry I missed that during the review process; I'll follow up with a more detailed post on-list about removing it.

bifurcation:
@jsha If we need to debate this in depth, we should do it on the list, but in case we can dispatch this quickly: I'm puzzled by your opposition to capability URLs.  When they're applied to POST-only URLs, they're strictly an improvement on the authentication / authz model.  It's only when they're applied to URLs where GETs are possible where there's some arguable weakening, in that you could choose GET+capability instead of POST+signature.  But if that's the case, then it seems like your issue is with allowing GETs at all, since once you allow GET, GET+capability is better than plain GET.

bifurcation:
Actually, @jsha, I'm going to go ahead and merge this, because it doesn't actually conflict with your proposed change in #457

jsha:
Actually, it does: The point is to not give people a false impression that URLs are unguessable when that is not a formal part of the security model. If it *is* a formal part of the security model, we need to analyze it properly.

PR #458, opened 2018-10-05
Revert "Randomize URLs in examples"

jsha:
Reverts ietf-wg-acme/acme#455


jsha:
Reverting #455 because it was merged with pending objections.

richsalz:
This should not have been merged without discussion on the list.

PR #459, opened 2018-10-05
Remove exception for certificate GET.

jsha:
This exception doesn't actually improve interop for STAR clients, and it introduces complexity into the security analysis of ACME. STAR can easily solve the problem by defining a new field on orders that provides an explicitly GETtable URL, with whatever security properties the STAR project needs.


richsalz:
Do not merge this without discussion on the mailing list.

bifurcation:

PR #460, opened 2018-10-06
Randomize URLs in examples

bifurcation:
The randomization of examples is independent of the questions over GETs.


bifurcation:

PR #462, opened 2018-10-07
Add a meta flag to indicate when GET requests for certificates are allowed

bifurcation:
This avoids the need for the client to guess.  Assumes that #459 will not be merged.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -840,6 +840,10 @@ externalAccountRequired (optional, boolean):
 new-account requests include an "externalAccountBinding" field associating the
 new account with an external account.
 
+certificateGET (optional, boolean):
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -840,6 +840,10 @@ externalAccountRequired (optional, boolean):
 new-account requests include an "externalAccountBinding" field associating the
 new account with an external account.
 
+certificateGET (optional, boolean):
+: If this field is present and set to "true", then the server allows GET
```
martinthomson: the server, or the resource?

bifurcation: The server.  This is supposed to apply to all certificate resources.

bifurcation:
Closing, since we're going to land #459 removing GET for certificates.

PR #463, opened 2018-10-09
Tweak security concerns for URL correlations.

jsha:
- Fix a misuse of the term "capability URL."
- Change the example. Hiding the number of accounts was not a great
example, because even with randomized account URLs, an attacker could
probe the space of URLs to estimate their density.


bifurcation:

PR #465, opened 2018-10-13
Fix Typo

linusaurusrex:


linusaurusrex:
Change "applicaiton" to "application"

richsalz:
With any luck we'll have to catch this in the RFC editor queue; hopefully there won't be another draft.

PR #467, opened 2018-11-03
Minor fixes in responses to pre-RFC-Editor review with EKR

bifurcation:
EKR and I did a pass through the IESG comments, and found a couple minor things that were missed.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -3384,14 +3384,13 @@ account holder could take within the scope of ACME:
 
 For this reason, it is RECOMMENDED that account key pairs be used for no other
 purpose besides ACME authentication.  For example, the public key of an account
-key pair SHOULD NOT be included in a certificate.  ACME clients MUST NOT reuse 
+key pair MUST NOT be included in a certificate.  ACME clients MUST NOT reuse
 the same account key for multiple accounts, and MUST NOT allow account key 
-roll-over to a previously-used account key.  ACME servers SHOULD reject a 
-new-account request using an account key already associated with an account 
-on the server.  
+roll-over to a previously-used account key.  ACME servers MUST NOT create a new
+account using an account key already associated with an account on the server.
 
 ACME clients and servers
-SHOULD verify that a CSR submitted in a finalize request does not contain a
+MUST verify that a CSR submitted in a finalize request does not contain a
 public key for any known account key pair.  In particular, when a server
 receives a finalize request, it MUST verify that the public key in a CSR is not
 the same as the public key of the account key pair used to authenticate that
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -226,7 +226,7 @@ from the client.
 # Protocol Overview
 
 ACME allows a client to request certificate management actions using a set of
-JavaScript Object Notation (JSON) messages carried over HTTPS.
+JavaScript Object Notation (JSON) messages carried over HTTPS {{!RFC7159}} {{!RFC2818}}.
 Issuance using ACME resembles a traditional CA's issuance process, in which a user creates an account,
 requests a certificate, and proves control of the domain(s) in that certificate in
 order for the CA to issue the requested certificate.
@@ -509,7 +509,8 @@ any signed request from the client to carry such a nonce.
 An ACME server provides nonces to clients using the HTTP Replay-Nonce header field,
 as specified in {{replay-nonce}} below.  The server MUST include a Replay-Nonce
 header field in every successful response to a POST request and SHOULD provide
-it in error responses as well.
+it in error responses as well.  Servers SHOULD use globally scoped nonces, so that
+a nonce, once issued, will be accepted on any HTTP request.
 
 Every JWS sent by an ACME client MUST include, in its protected header, the
 "nonce" header parameter, with contents as defined in
@@ -1276,12 +1277,11 @@ The server MUST include a Cache-Control header field with the "no-store"
 directive in responses for the new-nonce resource, in order to prevent
 caching of this resource.
 
-## Account Creation
+## Account Management 
 
 A client creates a new account with the server by sending a POST request to the
 server's new-account URL.  The body of the request is a stub account object
-optionally containing the "contact" and "termsOfServiceAgreed" fields, and
-optionally the "onlyReturnExisting" and "externalAccountBinding" fields.
+containing some subset of the following fields:
 
 contact (optional, array of string):
 : Same meaning as the corresponding server field defined in {{account-objects}}
@@ -1360,7 +1360,8 @@ JWS (i.e., the "jwk" element of the JWS header) to authenticate future requests
 from the account.  The server returns this account object in a 201 (Created)
 response, with the account URL in a Location header field. The account URL is
 used as the "kid" value in the JWS authenticating subsequent requests by this
-account (See {{request-authentication}}).
+account (see {{request-authentication}}).  The account URL is also used for
+requests for management actions on this account, as described below.
 
 ~~~~~~~~~~
 HTTP/1.1 201 Created
@@ -1817,7 +1818,7 @@ The CSR encodes the client's requests with regard to the content of the
 certificate to be issued.  The CSR MUST indicate the exact same set of requested
 identifiers as the initial new-order request.  Identifiers of type "dns" MUST appear either in the commonName portion
 of the requested subject name, or in an extensionRequest attribute {{!RFC2985}}
-requesting a subjectAltName extension.  (These identifiers may appear
+requesting a subjectAltName extension, or both.  (These identifiers may appear
 in any sort order.)  Specifications that define
 new identifier types must specify where in the certificate signing
 request these
@@ -2036,7 +2037,7 @@ in DER format. Server support for alternate formats is OPTIONAL. For
 formats that can only express a single certificate, the server SHOULD
 provide one or more `Link: rel="up"` header fields pointing to an issuer or
 issuers so that ACME clients can build a certificate chain as defined
-in TLS {{!RFC8446}}.
+in TLS (see Section 4.4.2 of {{!RFC8446}}).
 
 ## Identifier Authorization
 
@@ -2823,19 +2824,22 @@ Template:
 * Field name: The string to be used as a field name in the JSON object
 * Field type: The type of value to be provided, e.g., string, boolean, array of
   string
-* Client configurable: Boolean indicating whether the server should accept
-  values provided by the client
+* Requests: Either the value "none" or a list of types of requests
+  where the field is allowed in a request object, taken from the
+  following values:
+  * "new" - Requests to the "newAccount" URL
+  * "account" - Requests to an account URL
 * Reference: Where this field is defined
 
 Initial contents: The fields and descriptions defined in {{account-objects}}.
 
-| Field Name               | Field Type      | Configurable | Reference |
+| Field Name               | Field Type      | Requests     | Reference |
 |:-------------------------|:----------------|:-------------|:----------|
-| status                   | string          | false        | RFC XXXX  |
-| contact                  | array of string | true         | RFC XXXX  |
-| externalAccountBinding   | object          | true         | RFC XXXX  |
-| termsOfServiceAgreed     | boolean         | true         | RFC XXXX  |
-| orders                   | string          | false        | RFC XXXX  |
+| status                   | string          | new, account | RFC XXXX  |
+| contact                  | array of string | new, account | RFC XXXX  |
+| externalAccountBinding   | object          | new          | RFC XXXX  |
+| termsOfServiceAgreed     | boolean         | new          | RFC XXXX  |
+| orders                   | string          | none         | RFC XXXX  |
 
 \[\[ RFC EDITOR: Please replace XXXX above with the RFC number assigned to this
 document ]]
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -3380,7 +3384,13 @@ account holder could take within the scope of ACME:
 
 For this reason, it is RECOMMENDED that account key pairs be used for no other
 purpose besides ACME authentication.  For example, the public key of an account
-key pair SHOULD NOT be included in a certificate.  ACME clients and servers
+key pair SHOULD NOT be included in a certificate.  ACME clients MUST NOT reuse 
+the same account key for multiple accounts, and MUST NOT allow account key 
+roll-over to a previously-used account key.  ACME servers SHOULD reject a
```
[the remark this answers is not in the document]
felixfontein: I thought a new-account request with a public key already associated with an account is *the* way to retrieve the account URI, given the public key? I think neither SHOULD nor MUST is a good idea because of that.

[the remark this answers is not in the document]
felixfontein: I guess the sentence should be "ACME servers MUST NOT create a new account using an account key already associated with an account on the server.", without mentioning the `new-account` endpoint.

draft-ietf-acme-acme.md
```diff
@@ -1276,12 +1277,11 @@ The server MUST include a Cache-Control header field with the "no-store"
 directive in responses for the new-nonce resource, in order to prevent
 caching of this resource.
 
-## Account Creation
+## Account Management 
 
 A client creates a new account with the server by sending a POST request to the
```
felixfontein: Shouldn't it better be "... creates a new account, or manages an existing account, with the server by ..." (or something like that)?

bifurcation: The start of management is creation :)

draft-ietf-acme-acme.md
```diff
@@ -3380,8 +3384,13 @@ account holder could take within the scope of ACME:
 
 For this reason, it is RECOMMENDED that account key pairs be used for no other
 purpose besides ACME authentication.  For example, the public key of an account
-key pair SHOULD NOT be included in a certificate.  ACME clients and servers
-SHOULD verify that a CSR submitted in a finalize request does not contain a
+key pair MUST NOT be included in a certificate.  ACME clients MUST NOT reuse
+the same account key for multiple accounts, and MUST NOT allow account key 
+roll-over to a previously-used account key.  ACME servers MUST NOT create a new
```
cpu: I think adding "MUST NOT allow account key roll-over to a previously-used account key" is overly prescriptive and under specified for this late in the game.

 Why isn't this restriction also mentioned in the key rollover section of the draft? The list of steps the server needs to take in https://tools.ietf.org/html/draft-ietf-acme-acme-16#section-7.3.5 mentions "Check that no account exists whose account key is the same as the key in the "jwk" header parameter of the inner JWS." but nothing about also verifying that the new key has never been used by the account previously.

What threat is this addressing that justifies the new complexity? Are we sure the security win is proportional? What error is returned if a client tries to roll-over to a key that has been used by the account in the past? How much key history does the server need to keep? What happens if an ACME account decides to roll-over to a new randomly generated keypair every minute? What should servers do to avoid that being abused?

draft-ietf-acme-acme.md
felixfontein: It is a bit strange if you first read "A client creates a new account [...] by sending a POST request ...", and then you read in the details that the client can also send a parameter `onlyReturnExisting`. I don't mind too much if it is kept like this, though :)

draft-ietf-acme-acme.md
cpu: > Note that the "MUST NOT allow account key roll-over to a previously-used account key" is a requirement on ACME clients rather

Ahhhh! I apologize. I totally missed that part.

I agree with you that it seems like it isn't likely to get implemented. I'm also dubious of the value if the server doesn't enforce this constraint as well, and as mentioned I think the server enforcing it has its own set of complications.

ekr: I read "MUST NOT allow" as a clunky way of saying "MUST NOT do" and given that the client probably generates the account key itself, this seems quite easy to enforce.

WRT the value if the server doesn't enforce it, I'm not following why that would be the case. We, for instance, often want clients to generate fresh DH shares, even though the server usually does not check that they have done so.

ekr: In this case, this would be a requirement on the combined system of user and client software.

ekr: I don't think that that's right. You usually have to read IETF specs with the mindset that everything on one side of the connection is a monolithic unit, and this is no exception

ekr:
No, I don't think that follows. What I'm saying is that the way you say
that is "the client".

draft-ietf-acme-acme.md
```diff
@@ -509,7 +509,8 @@ any signed request from the client to carry such a nonce.
 An ACME server provides nonces to clients using the HTTP Replay-Nonce header field,
 as specified in {{replay-nonce}} below.  The server MUST include a Replay-Nonce
 header field in every successful response to a POST request and SHOULD provide
-it in error responses as well.
+it in error responses as well.  Servers SHOULD use globally scoped nonces, so that
```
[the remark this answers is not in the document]
bifurcation: There's an interop concern here, in that clients need to know how they can use nonces.  I would probably make the slightly stronger argument that we should forbid servers from doing things that would break the obvious ways for clients to handle nonces (i.e., in one pool per client).

For example, a server could have a scheme where (1) a `newNonce`-issued nonce is good for talking to any resource, but (2) nonces issued from other resources can only be used with resources of the same class.  That would cause a lot of fail/newNonce/retry loops for a client doing the obvious thing.

TBH, incorporating client IP seems kind of on the edge, as it's not crazy to imagine that a client could change addresses during the course of an ACME issuance flow.

How about something like this:

> Servers SHOULD NOT scope nonces more finely than per-account.  In particular, servers MUST ensure that a nonce issued to a client by the `newNonce` resource can be used by the same client to perform any ACME request (within a reasonable period of time).

The latter part ensures that the fail/newNonce/retry loop works.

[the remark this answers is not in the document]
cpu: > There's an interop concern here, in that clients need to know how they can use nonces. 

I don't think that's true. Clients that send a nonce to a server from the wrong scope will get back a nonce rejection error that itself has a new nonce to use that the server will accept subsequently. The spec already says that a client SHOULD retry in this case. Some imagined (e.g. not implemented, not discussed on list by anyone) server nonce implementations like per-resource scoping might provoke more retries from clients but that seems like a fair trade-off for a hypothetical design choice and not a concern that justifies changing the draft text.

> How about something like this:
>> Servers SHOULD NOT scope nonces more finely than per-account.

I don't think the text should have a "SHOULD NOT" added that directly contradicts all of the existing server implementations I'm aware of, and certainly the largest production ACME implementation and its ecosystem of compatible clients.

At this stage of the process I really think that every change should have an extremely strong justification. This change contradicts all implementation experience we have and the rationale for it (e.g. supporting servers implementing per-resource nonce schemes) does not seem strong enough to meet the bar for merging.

draft-ietf-acme-acme.md
cpu: > I think the desired state here is that if the client or server has the state to determine that key reuse has occurred, then it must check and reject reuse. 

Do any clients meeting this criteria exist? Does the benefit of addressing this niche justify last-minute specification changes? I feel like its no in both cases.

If we absolutely have to land this text to make forward progress then I'm not strongly opposed but it doesn't seem practically useful to me.

jsha:
@ekr what do you think of "clients MUST NOT intentionally reuse keys"? The addition of "intentional" would hopefully make it clear that client software is not required to keep a copy of all past keys around to ensure it can never possibly reuse them.

ekr:
We don't usually treat software as having intent. I don't see that it's a problem to forbid the endpoint from doing it. That simply isn't a levy on some specific piece of client software to keep a list; it's a levy on the system of the client software and whatever is managing the keys.

Happy to have a call to discuss this.

jsha:
In the interest of making progress I'm going to drop my objection to the "clients MUST NOT reuse" clause. I still object to "servers SHOULD scope nonces globally." I think we should just drop that clause - it's a very significant change to add in post-post-post-WGLC, for something that hasn't been an issue in practice.

draft-ietf-acme-acme.md
bifurcation: Added a bit of intro text to make this nicer.

bifurcation:
@cpu @jsha @ekr I've updated the text around nonce scoping and key reuse in a way that I hope makes folks more comfortable.  For nonce reuse, the text is focused on the interop problem, i.e., making the "get a fresh nonce and retry" cycle work; otherwise, nonce scope is left unspecified.  For key reuse, I tried to be clearer about who is responsible for doing what under what circumstances.

The latest push also mostly fixes the XSS issue noted in #470.  There's still some residual risk that a keyAuthorization will look like something else, but at least now the client's response is required to be a keyAuthorization.

draft-ietf-acme-acme.md
```diff
@@ -3378,10 +3397,20 @@ account holder could take within the scope of ACME:
...
+certificate.  If an ACME client receives a request from a user for
+account creation or key roll-over using an account key that the
+client knows to be used elsewhere, then the client MUST return an
+error.  Clients that manage account keys on behalf of users SHOULD
+generate a fresh account key for every account creation or roll-over
+operation.  Note that given the requirements of
+{{#finding-an-account-url-given-a-key}}, servers will not create
+accounts with reused keys anyway.
+
+ACME clients and servers
```
cpu: I think a spurious newline slipped in here

draft-ietf-acme-acme.md
```diff
@@ -3460,3 +3490,4 @@ inception.
 This document draws on many concepts established by Eric Rescorla's "Automated
 Certificate Issuance Protocol" draft.  Martin Thomson provided helpful guidance
 in the use of HTTP.
+
```
cpu: ditto re: newline here

ekr:
I'm not really happy with this text. The requirement ought to be that you are forbidden from reusing, period. Happy to have a call to discuss, but as-is, I'm not OK with this.

draft-ietf-acme-acme.md
```diff
@@ -522,7 +522,9 @@ way as a value it had never issued.
 When a server rejects a request because its nonce value was unacceptable (or not
 present), it MUST provide HTTP status code 400 (Bad Request), and indicate the
 ACME error type "urn:ietf:params:acme:error:badNonce".  An error response with
-the "badNonce" error type MUST include a Replay-Nonce header with a fresh nonce.
+the "badNonce" error type MUST include a Replay-Nonce header with a
+fresh nonce that the server will accept in a retry of the original
+query (and possibly in other requests).
```
shred: I hope I'm not too late for discussion... But what is "and possibly in other requests" supposed to mean? Does it mean that a client MUST NOT or SHOULD NOT use the Replay-Nonce from the badNonce error for other requests?

I'm not happy with this idea. Until now, nonces were not bound to a certain type of request. A client could just remember the nonce it received from the last request, and use it for the next request. If I take this change literally, a client would now have to store multiple nonces, and find the one matching for a certain request. I see no advantage or increase of security in that.

Or did I just misunderstand it?

bifurcation: You've misunderstood a little, but in a totally understandable way :)  Right now, ACME doesn't say anything about how nonces are scoped -- in principle, the server could scope them per-client, per-request-type, etc.  There are valid reasons for servers to do some of these things, mainly related to managing server state.

My intent with this text was to specify the minimum necessary scope constraint for a client and server to interoperate.  Without this requirement, the worst case is that the server has some crazy nonce management strategy and client can never send a valid request.  The requirement for the nonce to be valid on retry assures that the worst case is that the client has to do every request twice.  Obviously, I hope that most servers will use more sensible strategies that don't cause this condition!

I agree that the current text is awkward, though.  Would it be helpful to replace that parenthetical with something like the following?

> Other than this constraint, ACME does not constrain how servers scope nonces.  Clients are MAY assume that nonces have broad scope, e.g., by having a single pool of nonces used for all requests.  However, when retrying in response to a "badNonce" error, the client MUST use the nonce provided in the error response.   Obviously, servers should scope nonces broadly enough that retries are not needed very often.

ekr: I think Richard's proposed text here is good and addresses the review comment. @richsalz, are you good with this? I think it's the last thing.

ekr:
@bifurcation: anyone else you need to clear this with or shall we issue the
next I-D and send it to the RFC Ed

cpu:
>  anyone else you need to clear this with or shall we issue the
next I-D and send it to the RFC Ed

@ekr at the minimum someone needs to update the document with @bifurcation's proposed text first. I think @jsha should also have a chance to review the changes made over the weekend.

draft-ietf-acme-acme.md
```diff
@@ -532,6 +532,15 @@ The precise method used to generate and track nonces is up to the server.  For
 example, the server could generate a random 128-bit value for each response,
 keep a list of issued nonces, and strike nonces from this list as they are used.
 
+Other than the constraint above with regard to nonces issued in
+"badNonce" responses, ACME does not constrain how servers
+scope nonces. Clients MAY assume that nonces have broad scope,
+e.g., by having a single pool of nonces used for all requests.
+However, when retrying in response to a "badNonce" error, the client
+MUST use the nonce provided in the error response. Obviously,
```
cpu: nit: I think "Obviously," isn't warranted. Just say "Servers should scope nonces broadly enough ..."

bifurcation:
Merging despite CI failure because the CI failure appears to be tooling-related.  I have verified locally that the document builds correctly.

PR #468, opened 2018-11-06
Fix leftover refs to GETs for Cert resources.

cpu:
Thanks to Felipe Gasper for [pointing these errors out on list](https://mailarchive.ietf.org/arch/msg/acme/-Tin2by6SFBe8RRTxTB_eOlfKHk).


bifurcation:
Merging despite spurious CI failure, since it seems pretty clear that the CI failure is spurious.

PR #474, opened 2018-12-18
Typo fix

shred:


richsalz:
I suggest adding it to the list of AUTH48 nits.  I'm sure the RFC editor will find it :

bifurcation:
This ended up getting addressed in #484

PR #475, opened 2018-12-20
Tighten key reuse requirements

bifurcation:
In #467, I forgot to make one change that had been privately agreed among the authors and EKR.  This PR just incorporates that change.


cpu:
@bifurcation I can't merge this with the :x: CI, can you do the honours?

bifurcation:
Yep.  For posterity: As with #467, merging this with CI failures because they seem to be tooling related and it builds for me locally.

PR #476, opened 2019-01-21
Fix typo "concatinates"

joernheissler:


bifurcation:
This got fixed in the meantime by #484

PR #477, opened 2019-01-22
TODO items to fix during AUTH48

bifurcation:
This PR is collecting changes to make sure we address during the [AUTH48](https://www.rfc-editor.org/pubprocess/auth48/) editing window.  Editorial clarifications only, please.


cpu:
Both #474 and #476 are also TODO's for AUTH48.

robstradling:
Section 7.4 says:
```
In such cases, the problem document returned by the server SHOULD use
error code "badCSR", and describe specific reasons the CSR was
rejected in its "details" field.
```
It should say `"detail"`, not `"details"`.

@bifurcation Would you mind fixing this typo in this PR?

bifurcation:
@robstradling Turns out the RFC editor caught that one (see #484), so I don't have to fix it here :)

jsha:
LGTM

PR #479, opened 2019-01-28
Some clarifications regarding Account Objects

robstradling:
From Section 7.3:
```
   The server MUST ignore any values provided in the "orders" fields in
   account bodies sent by the client, as well as any other fields that
   it does not recognize.  If new fields are specified in the future,
   the specification of those fields MUST describe whether they can be
   provided by the client.

   In general, the server MUST ignore any fields in the request object
   that it does not recognize.  In particular, it MUST NOT reflect
   unrecognized fields in the resulting account object.  This allows
   clients to detect when servers do not support an extension field.
```
"MUST ignore...in account bodies sent by the client...any other fields that it does not recognize" in the first paragraph seems to be unnecessarily repeated by "MUST ignore any fields in the request object that it does not recognize" in the second paragraph.  This PR merges these two paragraphs and discards the duplicated text.

Also, "account bodies" isn't used anywhere else in the document, so this PR changes it to "account objects".


jsha:
This looks reasonable to handle in AUTH48.

PR #482, opened 2019-02-19
Add 'badState' error.

robstradling:
draft-ietf-acme-acme-18 §7.4 has two conflicting statements regarding what an ACME server should do when it receives a request to finalize an order that is not in the "ready" state:
1. '_If the order indicated does not have status "ready" ... the problem document returned by the server SHOULD use error code "badCSR"_'.
2. '_A request to finalize an order will return the order to be finalized.  The client should begin polling the order by sending a POST-as-GET request to the order resource to obtain its current state.  The status of the order will indicate what action the client should take:_', where values other than "ready" are listed.

This PR proposes to resolve this conflict by requiring servers to return a new 'badState' error in response to a request to finalize an order that is not in the "ready" state.  The client should respond to a "badState" error by re-fetching the order object and examining the status field to determine what to do.


jsha:
I think a better name for this very specific error would be badOrderState or orderNotReady. That avoids potential confusion with bad states for authzs or challenges. Posted a proposed change in https://github.com/ietf-wg-acme/acme/pull/490.

PR #490, opened 2019-02-20
Fix some nits for AUTH48.

jsha:
- badState should be orderNotReady
- Use "wildcard domain name" consistently instead of "wildcard prefix"

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -621,7 +621,6 @@ in the "type" field (within the ACME URN namespace "urn:ietf:params:acme:error:"
 | badPublicKey            | The JWS was signed by a public key the server does not support                               |
 | badRevocationReason     | The revocation reason provided is not allowed by the server                                  |
 | badSignatureAlgorithm   | The JWS was signed with an algorithm the server does not support                             |
-| badState                | The request attempted to finalize an order that is not ready to be finalized                 |
 | caa                     | Certification Authority Authorization (CAA) records forbid the CA from issuing a certificate |
 | compound                | Specific error conditions are indicated in the "subproblems" array.                          |
 | connection              | The server could not connect to validation target                                            |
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1868,7 +1868,7 @@ client to submit a new finalize request with an amended CSR.
 
 A request to finalize an order will result in error if the order is not in the
 "ready" state.  In such cases, the server MUST return a 403 (Forbidden) error
-with a problem document of type "badState".  The client should then send a
+with a problem document of type "badOrderState".  The client should then send a
```
jdkasten: I assume this should be "orderNotReady"?

yoavnir: Indeed.

draft-ietf-acme-acme.md
```diff
@@ -1981,8 +1981,8 @@ Content-Type: application/jose+json
 
 Note that because the identifier in a pre-authorization request is
 the exact identifier to be included in the authorization object,
-pre-authorization cannot be used to authorize issuance with wildcard
-DNS identifiers.
+pre-authorization cannot be used to authorize issuance of certificates
+containing wildcard domain names.
```
yoavnir: This is a very minor nit, but is the term "wildcard domain name" common?  I thought that you could either have a "wildcard DNS identifier" in a certificate, or simply a "wildcard certificate" but that domain names were never wildcard.

jsha: For better or for worse, "wildcard domain name" is the term that wound up codified in the BRs:

> Wildcard Domain Name: A Domain Name consisting of a single asterisk character followed by a single full stop character (“*.”) followed by a Fully-Qualified Domain Name.

Since this is the most rigorous definition of a wildcard available, we're starting to standardize around it, e.g. also in 6844bis. It has the downside that a "wildcard domain name" is not actually a "domain name," but it's progress anyhow.

jsha: By the way, this is just harmonizing the language in different sections. In {{order-objects}} we talk about "wildcard domain names" and define them, but in {{pre-authorization}} we were talking about "wildcard DNS identifiers," which aren't defined anywhere but mean the same thing.

yoavnir: Cool

PR #493, opened 2019-02-20
Include the 'index' link relation in all example server responses that return a resource

robstradling:
Section 7.1 says:
```The "index" link relation is present on all resources other than the directory and indicates the URL of the directory.```

This PR adds an example "index" link relation to the example server responses that return a resource but that didn't already have it.

I'm interpreting "is present on all resources" to _not_ include responses that return a problem document.  Is that right?

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -932,6 +932,7 @@ relation indicating where further entries can be acquired.
 ~~~~~~~~~~
 HTTP/1.1 200 OK
 Content-Type: application/json
+Link: <https://example.com/acme/some-directory>;rel="index"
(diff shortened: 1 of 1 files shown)


bifurcation:
I agree with @cpu here -- an HTTP response reflects the status of a resource, even if that status is a problem

robstradling:
OK.  I've made that change.

robstradling:
@cpu Yeah, I'd noticed that and I'd considered switching the order for consistency.  But then I thought perhaps it was wrong to be unnecessarily consistent, because that might be taken as implying that the order of the HTTP response headers matters (which it doesn't).
I have no strong feelings though, so I'll switch them.

PR #496, opened 2019-02-20
Some more edits from thorough review on paper.

jsha:
- Some grammar fixes
- POST-as-GETs should be indicated as having a signature in flow
- Specify that the JWS profile we define in Section 6.2 is for ACME request bodies, to
  distinuish it from the profile used for External Account Binding. They
  have incompatible requirements: MUST NOT use MAC vs MUST use MAC.
- Clarify that Replay-Nonce is an HTTP header field since it's defined
  near JWS header parameters.
- Don't include "wildcard": false in examples. Section 7.1.4 requires
  that it MUST be absent when false.
- There was lingering language in the Section 7.5.1 about "updating"
  an authorization body in response to fields in a challenge POST.
  Since the challenge POST is now described explicitly as "an empty
  JSON body", this language didn't make sense. Fixed.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1534,7 +1534,7 @@ Content-Type: application/jose+json
     "url": "https://example.com/acme/new-account"
   }),
   "payload": base64url({
-    "contact": ["mailto:cert-admin@example.com"],
+    "contact": ["mailto:example@anonymous.invalid"],
     "termsOfServiceAgreed": true,
 
     "externalAccountBinding": {
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -112,7 +112,7 @@ certificates, a typical user experience is something like:
 
 * Generate a PKCS#10 {{!RFC2986}} Certificate Signing Request (CSR).
 * Cut and paste the CSR into a CA's web page.
-* Prove ownership of the domain by one of the following methods:
+* Prove ownership of the domain(s) in the CSR by one of the following methods:
    * Put a CA-provided challenge at a specific place on the web server.
    * Put a CA-provided challenge in a DNS record corresponding to the target
      domain.
@@ -185,7 +185,7 @@ deploying an HTTPS server using ACME, the experience would be something like thi
   having the ACME client perform some action(s) that can only be done
   with control of the domain name(s).
   For example, the CA might require a client requesting example.com
-  to provision DNS record under example.com or an HTTP resource
+  to provision a DNS record under example.com or an HTTP resource
   under http://example.com.
 * Once the CA is satisfied, it issues the certificate and the ACME client
   automatically downloads and installs it, potentially notifying the operator
@@ -299,7 +299,8 @@ certificate and make it available to the client.
 
                           <~~~~~~Await issuance~~~~~~>
 
-      POST-as-GET request           ------->
+      [POST-as-GET request]
+      Signature                     ------->
                                     <-------              Certificate
 
                 [] Information covered by request signatures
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1885,21 +1886,17 @@ action the client should take:
...
-  request.
-
 * "processing": The certificate is being issued. Send a POST-as-GET request after the
   time given in the Retry-After header field of the response, if
   any.
 
 * "valid": The server has issued the certificate and provisioned its URL to the
   "certificate" field of the order.  Download the certificate.
 
+* "pending": This status should not be returned after a successful finalize request.
```
jdkasten: L1878-L1879 references these status definitions when an "orderNotReady" failure is returned, so I believe fully defining each of the statuses is clearer.

**Note "The status of the order will indicate what action the client should take" is repeated twice within three lines so it may be helpful to additionally reword it, but I didn't come up with a quick fix.

jsha: Good point, thanks. Reverted this portion of the diff.

cpu: @jsha Can you also update the PR desc to remove mention of the removed diff?

jsha: Done.

jsha:
Per offline review, I added back in a sentence that contained a MUST:

> The server MUST ignore any fields in the response object that are not specified as response fields for this type of challenge.  Note that the challenges in this document do not define any response fields, but future specifications might define them.

draft-ietf-acme-acme.md
```diff
@@ -398,7 +399,7 @@ private key unless otherwise specified.  The server MUST verify the JWS before
 processing the request. Encapsulating request bodies in JWS provides
 authentication of requests.
 
-JWS objects sent in ACME requests MUST meet the following additional criteria:
+JWS objects sent as the body of ACME requests MUST meet the following additional criteria:
```
bifurcation: This doesn't quite parse.  How about we change to the singular: "A JWS object sent as the body of an ACME request MUST..."

draft-ietf-acme-acme.md
```diff
@@ -2178,10 +2177,11 @@ Content-Type: application/jose+json
 }
 ~~~~~~~~~~
 
-The server updates the authorization document by updating its representation of
-the challenge with the response object provided by the client.  The server MUST
+The server MUST
```
bifurcation: This seems like an unnecessarily heavy cut.  How about if we change "with" to "based on" -- the server takes the client's response, does some stuff, and updates its state.

draft-ietf-acme-acme.md
```diff
@@ -2193,10 +2193,11 @@ that have been provisioned to a web server.
 The server is said to "finalize" the authorization when it has completed
 one of the validations.  This is done by assigning the authorization a status of "valid"
 or "invalid", corresponding to whether it considers the account authorized
-for the identifier. If the final state is "valid", then the server MUST include
-an "expires" field. When finalizing an authorization, the server MAY remove
-challenges other than the one that was completed, and it may modify the "expires"
-field. The server SHOULD NOT remove challenges with status "invalid".
+for the identifier. If the final state is "valid", then the server MUST set
+an "expires" field, and MAY update the "expires" field if one was already present.
```
bifurcation: I'm not thrilled with adding a MAY here.  I think it's clear enough that an earlier "expires" field can be overridden.

draft-ietf-acme-acme.md
jsha: I decided to let the whole paragraph stand and just add the clarifying note.

PR #498, opened 2019-02-21
Normalize examples

jsha:
Since `example.com` is consistently used to represent the ACME server, we should always use `www.example.org` to represent the ACME client. Whenever there's a single hostname we prefer the `www` variant because the prefix makes it easier to quickly distinguish from `example.com`. When there are two hostnames we use both `www.example.org` and `example.org`.

Since we already used one order ID consistently through the doc, I made sure that its identifiers, notBefore, and notAfter were always the same. I tweaked the "expires" so it would be in a reasonable range (not precisely equal to the notBefore). Note that the "pending" version of the order has a slightly different expires than the "valid" order.

One version of the finalized order had a cert ID of jWCdfHVGY2M and another had mAt3xBGaobw. Harmonized on mAt3xBGaobw since it was also used in the certificate retrieval example.

Also fixed an accidentally duplicated nonce.

### DIFF
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1022,7 +1022,7 @@ certificate (optional, string):
 
   "finalize": "https://example.com/acme/order/TOlocE8rfgo/finalize",
 
-  "certificate": "https://example.com/acme/cert/jWCdfHVGY2M"
+  "certificate": "https://example.com/acme/cert/mAt3xBGaobw"
 }
 ~~~~~~~~~~
 
--- a/draft-ietf-acme-acme.md
+++ b/draft-ietf-acme-acme.md
@@ -1005,7 +1005,7 @@ certificate (optional, string):
 ~~~~~~~~~~
 {
   "status": "valid",
-  "expires": "2015-03-01T14:09:07.99Z",
+  "expires": "2016-01-20T14:09:07.99Z",
 
   "identifiers": [
     { "type": "dns", "value": "www.example.org" }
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-acme.md
```diff
@@ -1005,11 +1005,11 @@ certificate (optional, string):
 ~~~~~~~~~~
 {
   "status": "valid",
-  "expires": "2015-03-01T14:09:07.99Z",
+  "expires": "2016-01-20T14:09:07.99Z",
 
   "identifiers": [
-    { "type": "dns", "value": "example.com" },
-    { "type": "dns", "value": "www.example.com" }
+    { "type": "dns", "value": "www.example.org" }
```
cpu:
```suggestion
    { "type": "dns", "value": "www.example.org" },
```

draft-ietf-acme-acme.md
```diff
@@ -1005,11 +1005,11 @@ certificate (optional, string):
...
+    { "type": "dns", "value": "example.org" },
```
cpu:
```suggestion
    { "type": "dns", "value": "example.org" }
```

draft-ietf-acme-acme.md
```diff
@@ -1791,17 +1794,19 @@ Location: https://example.com/acme/order/TOlocE8rfgo
...
-  "expires": "2016-01-01T00:00:00Z",
+  "expires": "2016-01-05T14:09:07.99Z",
 
   "notBefore": "2016-01-01T00:00:00Z",
   "notAfter": "2016-01-08T00:00:00Z",
 
   "identifiers": [
-    { "type": "dns", "value": "example.com" },
+    { "type": "dns", "value": "www.example.org" },
+    { "type": "dns", "value": "example.org" },
```
cpu:
```suggestion
    { "type": "dns", "value": "example.org" }
```

draft-ietf-acme-acme.md
```diff
@@ -1908,14 +1913,14 @@ Location: https://example.com/acme/order/TOlocE8rfgo
...
 
-  "notBefore": "2015-12-31T00:17:00.00-09:00",
-  "notAfter": "2015-12-31T00:17:00.00-09:00",
+  "notBefore": "2016-01-01T00:00:00Z",
+  "notAfter": "2016-01-08T00:00:00Z",
 
   "identifiers": [
-    { "type": "dns", "value": "example.com" },
-    { "type": "dns", "value": "www.example.com" }
+    { "type": "dns", "value": "www.example.org" }
```
cpu:
```suggestion
    { "type": "dns", "value": "www.example.org" },
```

draft-ietf-acme-acme.md
```diff
@@ -1908,14 +1913,14 @@ Location: https://example.com/acme/order/TOlocE8rfgo
...
+    { "type": "dns", "value": "example.org" },
```
cpu:
```suggestion
    { "type": "dns", "value": "example.org" }
```

PR #500, opened 2019-02-21
Reconciliation

bifurcation:
This is (hopefully) the final reconciliation between the XML generated from the Markdown source and XML provided by the RFC Editor.  `rfc8555.src-clean.xml` will be provided to the authors for final review, and then to the RFC editor.


bifurcation:
Merging myself because this is just technical transformations of the source.

